PR #3, opened 2020-12-23
[ARTS-258] Add github action to build example-service

itye-msft:
this action will trigger a ci on new PR and on merge


tamirkamara (approved):

marrobi:
@itye-msft have you seen the work @shajidissa and @mdp0 have done here https://github.com/NDPH-ARTS/trial-config-service/blob/main/.github/workflows/pr-test.yml? The build and tests are happening within Docker. 

Should probably stick with one approach. I like the Docker approach, as I can build locally with no prerequisites, but happy to be convinced otherwise!

itye-msft:
> @itye-msft have you seen the work @shajidissa and @mdp0 have done here https://github.com/NDPH-ARTS/trial-config-service/blob/main/.github/workflows/pr-test.yml? The build and tests are happening within Docker.
> 
> Should probably stick with one approach. I like the Docker approach, as I can build locally with no prerequisites, but happy to be convinced otherwise!

I havn't seen it, and it looks good, aside of the the problem, that this yaml file should build all the projects inside the repo...

marrobi:
Hmm. Is each microservice not going to be built independently/have its own pipeline and tests? 

Guess could build the entire repo and then take the jar files, but could be building (and testing) lots that's isn't needed.

itye-msft:
> Hmm. Is each microservice not going to be built independently/have its own pipeline and tests?
> 
> Guess could build the entire repo and then take the jar files, but could be building (and testing) lots that's isn't needed.

The CI doesn't know all microservices are placed in the same repo. As far as it cares, there was a change which triggers a CI and now it builds... everything

marrobi:
As discussed, if this is just a placeholder until other pipelines are added then that's fine.

Suggest  a repo wide CI approach is agreed and this removed when resulting pipeline(s) are added.

.github/workflows/docker-build-push.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GHCR_LOGIN }}
+      - name: Build and push
+        run: |
+          docker build -t ghcr.io/ndph-arts/trial-config-service:${{ github.sha }} .
+          docker push ghcr.io/ndph-arts/trial-config-service:${{ github.sha }}
```
LizaShak, line 34: shouldn't it  be sample-service?

.github/workflows/docker-build-push.yml
```diff
@@ -0,0 +1,34 @@
+name: Docker build and push
```
LizaShak, line 1: I can't comment on the file, but we need to add the .DS_Store to gitignore file

LizaShak (changes requested):

LizaShak (approved):

PR #4, opened 2020-12-23
[ARTS-255] support VS Code devcontainer 

balteravishay:
This PR adds support for VS code [devcontainers](https://code.visualstudio.com/docs/remote/containers).


tamirkamara (approved):

eladiw (approved):

LizaShak (approved):

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,26 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.154.0/containers/java/.devcontainer/base.Dockerfile
+
+# [Choice] Java version: 11, 15
+ARG VARIANT="15"
```
itye-msft, line 4: Why is 15 an option? I think we are going only with 11, don't we?

itye-msft (approved):
Please see my comment

.devcontainer/Dockerfile
shajidissa: +1

shajidissa (approved):
All ok

PR #6, opened 2020-12-27
[ARTS-177] Practitioner Service implementation

balteravishay:
Implementing save a person to practitioner endpoint of a FHIR store in practitioner service.
the current implementation verifies that the values of prefix, given name and family name are not empty or null, until further validation is performed as part of ARTS-178.


practitioner-service/pom.xml
```diff
@@ -3,21 +3,22 @@
...
   <modelVersion>4.0.0</modelVersion>  
   <parent> 
-    <groupId>org.springframework.boot</groupId>  
-    <artifactId>spring-boot-starter-parent</artifactId>  
-    <version>2.3.6.RELEASE</version>  
-    <relativePath/>  
+    <groupId>uk.ac.ox.ctsu</groupId>  
+    <artifactId>java-base-app</artifactId>  
+    <version>0.0.1-SNAPSHOT</version>  
+    <relativePath>../</relativePath>
```
tamirkamara, line 9: Since you fixed the inheritance from the parent POM, then probably should also remove all dependencies here that come from the parent. right?

practitioner-service/pom.xml
```diff
@@ -3,21 +3,22 @@
...
-    <version>2.3.6.RELEASE</version>  
-    <relativePath/>  
+    <groupId>uk.ac.ox.ctsu</groupId>  
+    <artifactId>java-base-app</artifactId>  
+    <version>0.0.1-SNAPSHOT</version>  
+    <relativePath>../</relativePath>  
     <!-- lookup parent from repository --> 
   </parent>  
   <groupId>uk.ac.ox.ndph.arts</groupId>
   <artifactId>practitioner_service</artifactId>
```
tamirkamara, line 13: probably the convention here is practitioner-service (not 100% sure)

practitioner-service/pom.xml
```diff
@@ -3,21 +3,22 @@
...
+    <artifactId>java-base-app</artifactId>  
+    <version>0.0.1-SNAPSHOT</version>  
+    <relativePath>../</relativePath>  
     <!-- lookup parent from repository --> 
   </parent>  
   <groupId>uk.ac.ox.ndph.arts</groupId>
   <artifactId>practitioner_service</artifactId>  
   <version>0.0.1-SNAPSHOT</version>  
-  <name>practitioner-service</name>  
+  <name>practitioner_service</name>
```
tamirkamara, line 15: This is a friendly name, so code be:
```suggestion
  <name>Practitioner Service</name>  
```

practitioner-service/src/main/java/uk/ac/ox/ndph/arts/practitioner_service/repository/IFhirRepository.java
```diff
@@ -0,0 +1,19 @@
+package uk.ac.ox.ndph.arts.practitioner_service.repository;
+
+import org.hl7.fhir.r4.model.Practitioner;
+import uk.ac.ox.ndph.arts.practitioner_service.exception.HttpStatusException;
+
+/**
+* Interface for a FHIR entity repository
+*/
+public interface IFhirRepository {
```
tamirkamara, line 9: In Java you usually do use the **I** interface notation (guessing - since the most common IDEs like Eclipse and IntelliJ have an icon for it).
This applies to all interfaces in this PR.

practitioner-service/src/main/java/uk/ac/ox/ndph/arts/practitioner_service/service/PersonService.java
```diff
@@ -0,0 +1,48 @@
...
+import uk.ac.ox.ndph.arts.practitioner_service.repository.IFhirRepository;
+import uk.ac.ox.ndph.arts.practitioner_service.exception.HttpStatusException;
+import uk.ac.ox.ndph.arts.practitioner_service.exception.ArgumentException;
+
+/**
+* Implement an IEntityService.
+* validation is for empty values on all fields.
+*/
+@Service
+public class PersonService implements IEntityService {
```
tamirkamara, line 20: Any special reason this is Person and not Practitioner?

practitioner-service/src/main/java/uk/ac/ox/ndph/arts/practitioner_service/service/PersonService.java
```diff
@@ -0,0 +1,48 @@
...
+    @Autowired
+    public PersonService(IFhirRepository fhirRepository) {
+        this.fhirRepository = fhirRepository;
+    }
+
+    public String savePerson(Person person) throws HttpStatusException{
+        if (person.getPrefix() == null || person.getPrefix().isEmpty() || person.getGivenName() == null
+                || person.getGivenName().isEmpty() || person.getFamilyName() == null
+                || person.getFamilyName().isEmpty()) {
+            throw new ArgumentException("values cannot be null or empty");
```
tamirkamara, line 33: 1. Should these be broken to return a more specific error?
2. I think you should do isBlank (and not is empty) to check whitespaces too.
3. Perhaps, create a method to do both null and blank checks or even go for a package line StringUtils (in Apache commons)

p.s. another approach could be to do this verification on the entity itself and not allow assignments of the instance properties if null or blank.

practitioner-service/src/main/resources/application.properties
```diff
@@ -0,0 +1 @@
+fhir.uri=https://fhirbalter.azurewebsites.net
```
tamirkamara, line 1: should this be here?

tamirkamara (changes requested):

practitioner-service/pom.xml
balteravishay: inheritance of POM is not fixed as part of this PR, and the projects still do not work properly in any IDE. 
The POM structure is part of another user story, these are just initial settings which were wrong.

practitioner-service/pom.xml
balteravishay: "-" in namespace doesn't compile.

practitioner-service/src/main/java/uk/ac/ox/ndph/arts/practitioner_service/service/PersonService.java
balteravishay: wasn't sure if this is a practitioner for the rest of the system, or is it a person. the user story wasn't clear on it. do you think it's model should be practitioner?

practitioner-service/src/main/java/uk/ac/ox/ndph/arts/practitioner_service/service/PersonService.java
balteravishay: regarding the p.s. comment. the Person DTO is the input to the API, it is generated by spring so I couldn't put validations there. additionally, the validation logic in US 178 should come here as well.

practitioner-service/src/main/java/uk/ac/ox/ndph/arts/practitioner_service/model/Person.java
```diff
@@ -0,0 +1,25 @@
...
+@Getter
+@Setter
+@NoArgsConstructor
+public class Person implements Serializable {
+
+    private String prefix;
+    private String givenName;
+    private String familyName;
+
+    public Person(String prefix, String givenName, String familyName) {
```
SharonHart, line 20: This is actually an @AllArgsConstructor, so can be dropped as well. Is it required for constructing the object from the application or for Jackson serde operations?

practitioner-service/src/main/java/uk/ac/ox/ndph/arts/practitioner_service/exception/ArgumentException.java
```diff
@@ -0,0 +1,14 @@
+package uk.ac.ox.ndph.arts.practitioner_service.exception;
+
+import org.springframework.http.HttpStatus;
+
+public class ArgumentException extends HttpStatusException{
```
SharonHart, line 5: For all the exceptions that should be mapped to custom http status codes, you can use spring 'ResponseStatus' annotation: @ResponseStatus(value = HttpStatus. BAD_REQUEST)
Spring's REST service will automatically resolve the exception and map to a response with the relevant http status code.   
This can also simplify each exception class, that now can extend RuntimeException directly.

practitioner-service/src/test/java/uk/ac/ox/ndph/arts/practitioner_service/service/PersonServiceTests.java
```diff
@@ -0,0 +1,77 @@
...
+import org.mockito.ArgumentCaptor;
+import org.junit.jupiter.api.Assertions;
+import uk.ac.ox.ndph.arts.practitioner_service.model.Person;
+import uk.ac.ox.ndph.arts.practitioner_service.service.EntityService;
+import uk.ac.ox.ndph.arts.practitioner_service.repository.FhirRepository;
+import uk.ac.ox.ndph.arts.practitioner_service.exception.HttpStatusException;
+import uk.ac.ox.ndph.arts.practitioner_service.exception.RestException;
+import uk.ac.ox.ndph.arts.practitioner_service.exception.ArgumentException;
+
+@SpringBootTest
```
SharonHart, line 22: Not sure that this is required here since its not bootstrapping a spring application, just unit tests the service. 
You can also make the tests shorter by using the Mock annotations as class fields, like declaring the mock once and using it in all tests, each tests resets the stubbing you already did on the mocks:

@Mock
private FhirRepository  fhirRepository

practitioner-service/src/main/java/uk/ac/ox/ndph/arts/practitioner_service/repository/AzureFhirRepository.java
```diff
@@ -0,0 +1,81 @@
...
+    private final Logger logger = LoggerFactory.getLogger(AzureFhirRepository.class);
+
+    @Value("${fhir.uri}")
+    private String fhirUri = "";
+
+    public AzureFhirRepository() {
+        fhirContext = FhirContext.forR4();
+    }
+
+    public String savePractitioner(Practitioner practitioner) throws HttpStatusException {
```
SharonHart, line 39: From the same reason, that spring resolves the exceptions itself, you don't have to declare throws. The practice of rethrowing checked exceptions as runtime as you did below is actually good, it is just recommended to avoid the method declaration as much as possible since eventually you find yourself with many methods that have the declaration and many clients that needs to add the try-catch block.

tamirkamara (approved):
Nice.

SharonHart (approved):

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/repository/AzureFhirRepository.java
```diff
@@ -0,0 +1,84 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+import ca.uhn.fhir.context.FhirContext;
+import ca.uhn.fhir.rest.server.exceptions.BaseServerResponseException;
+import ca.uhn.fhir.util.BundleUtil;
+import uk.ac.ox.ndph.mts.practitioner_service.exception.RestException;
+
+/**
+ * Implement FhirRepository interface using HAPI client sdk and backed up by an Azure
```
eladiw, line 20: can you explain why Hapi client? or where do I see that is Hapi

also isnt the name should be 'backed up by an Azure FHIR server'? (instead of store?)

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/service/PractitionerService.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    private void validateArgument(String value, String argumentName){
+        if (value == null || value.isBlank()){
+            throw new ArgumentException(String.format(ERROR_MESSAGE, argumentName));
+        }
+    }
+
+    private org.hl7.fhir.r4.model.Practitioner toFhirPractitioner(Practitioner practitioner) {
+        org.hl7.fhir.r4.model.Practitioner fhirPractitioner = new org.hl7.fhir.r4.model.Practitioner();
```
eladiw, line 46: to my understanding only  adding new practiotioner is supported. updating one is currently  not supported? right?
since this method is not getting a uuid, but always generating one

eladiw (changes requested):
some questions and comments

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/model/Practitioner.java
```diff
@@ -0,0 +1,22 @@
...
+
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+public class Practitioner implements Serializable {
+
+    private String prefix;
+    private String givenName;
+    private String familyName;
```
LizaShak, line 20: I am not sure , but wanted to check anyway, if I look at Practitioner name which is a HumanName FHIR datatype (https://www.hl7.org/fhir/datatypes.html#HumanName) then givenName and prefix are arrays, am I right?

shajidissa (commented):
All ok

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/service/PractitionerService.java
eladiw: talk f2f. to my understanding currently update is not supported

eladiw (approved):

LizaShak (approved):

PR #7, opened 2020-12-28
Added parent pom

itye-msft:
The sample service is referencing the Parent pom + namespace changes to `ac.ox.ndph.mts`


sample-service/src/main/java/uk/ac/ox/ndph/mts/sampleservice/DemoApp.java
```diff
@@ -1,6 +1,6 @@
-package com.microsoft.baseapp;
+package uk.ac.ox.ndph.mts.sampleservice;
```
tamirkamara, line 1: I believe this should be sample_service. isn't it?

sample-service/Dockerfile
```diff
@@ -1,13 +1,14 @@
 FROM maven:3-openjdk-11 as builder
+COPY ./sample-service/pom.xml sample-service/pom.xml
```
tamirkamara, line 2: Can you get the service name from the outside so that this file can be generic?

.github/workflows/docker-build-push.yml
```diff
@@ -20,10 +20,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Build & test
-        working-directory: "${{ matrix.service_name }}"
+        #working-directory: "${{ matrix.service_name }}"
```
LizaShak, line 23: Is that what caused the problem yesterday? If we don't need that, and it causes issues... Maybe it is better to delete it then comment out?

pom.xml
```diff
@@ -29,7 +44,7 @@
           <artifactId>spring-boot-starter-logging</artifactId>
         </exclusion>
       </exclusions>
-    </dependency> 
+    </dependency>
```
LizaShak, line 47: redundant space, I know it is not part of this PR ... But we must have a story that runs linter/checkstyle as part of the CI

LizaShak (changes requested):

.github/workflows/docker-build-push.yml
itye-msft: done. pls resolve

sample-service/Dockerfile
itye-msft: @tamirkamara  yes, done, pls resolve.

sample-service/src/main/java/uk/ac/ox/ndph/mts/sampleservice/DemoApp.java
itye-msft: @tamirkamara  yes, pls resolve

pom.xml
itye-msft: done

LizaShak (approved):

tamirkamara (approved):

PR #8, opened 2020-12-29
Feature/trial config service

shajidissa:
POM has spring cloud dependancies commented out to make it work in development. 

Also the following properties are required:

    -Djdbc.driver=**db driver class name** 
    -Djdbc.url=< **arts db path** > 
    -Dserver.port=**http port**

Should these be in the key-vault ?


trial-config-service/README.md
```diff
@@ -0,0 +1,14 @@
+# trial-config-service
+
+The following properties are required:
+
+    -Djdbc.driver=**db driver class name**
```
itye-msft, line 5: I recommend to obtain credentials from KeyVault. You can find an example in the `sample-service`

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trialconfigservice/config/SecurityConfig.java
```diff
@@ -0,0 +1,33 @@
...
+
+
+
+    @Override
+    protected void configure(HttpSecurity http) throws Exception {
+        http.authorizeRequests()
+            .antMatchers(HttpMethod.OPTIONS,  "/trial-config/**").permitAll()
+            .anyRequest().authenticated()
+            .and().oauth2ResourceServer(oauth2 -> oauth2.jwt()
+                /*.jwtAuthenticationConverter(getCustomJwtAuthConverter())*/);// not yet
```
itye-msft, line 20: Please add a TODO comment with a user story number for any technical debt or future work to be done.

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trialconfigservice/config/WebConfig.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.util.Arrays.asList;
+
+@Configuration
+public class WebConfig implements WebMvcConfigurer {
+    @Value("${spring.security.oauth2.resourceserver.opaquetoken.client-secret}")
+    private String CLIENT_SECRET;
```
itye-msft, line 27: There should be a naming convention. Usually Capital case is used for static members.
This is confusing.
Why can't it be `clientSecret` as the standard naming convention? It can be other option, but for the sake of a unified code with other developers, I suggest to align to the standard, or to suggest a new naming convention

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trialconfigservice/config/WebConfig.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Value("${spring.security.oauth2.resourceserver.opaquetoken.client-id}")
+    private String CLIENT_ID;
+
+    @Value("${swagger.authserver.url}")
+    private String AUTH_SERVER;
+
+    private List<AuthorizationScope> authorizationScopeList = new ArrayList<>();
```
itye-msft, line 35: It's more common to initialise members in the constructor.

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trialconfigservice/config/WebConfig.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Bean
+    public Docket api() {
+        authorizationScopeList.add(new AuthorizationScope("api://mts-apps/admin", "admin scope"));
```
itye-msft, line 39: Please extract hard coded strings into static members, or even better, to environment variables, or other sort or configurable settings.

trial-config-service/src/main/resources/application.properties
```diff
@@ -0,0 +1,20 @@
...
+spring.hateoas.use-hal-as-default-json-media-type=false
+
+
+# Log properties in application.properties
+#logging.level.org.springframework.security=DEBUG
+
+# Specifies your Active Directory ID:
+spring.security.oauth2.resourceserver.jwt.issuer-uri=https://sts.windows.net/99804659-431f-48fa-84c1-65c9609de05b/
+spring.security.oauth2.resourceserver.opaquetoken.client-id=901b377a-a4c0-4a72-b390-294cef5c0c74
+spring.security.oauth2.resourceserver.opaquetoken.client-secret=wq99fHb3Q9Vg_LfsZ_pNEG6~XR_h8um6u4
```
itye-msft, line 15: This secret is now compromised. Please remove from the code and create a new secret.
It's better to create a managed identity instead of committing secrets into a public repo.

trial-config-service/src/test/java/uk/ac/ox/ndph/mts/trialconfigservice/controller/TrialConfigControllerTest.java
```diff
@@ -0,0 +1,74 @@
...
+    @Mock
+    WebConfig webConfig;
+
+    @BeforeEach
+    void setUp() {
+        trialConfigController = new TrialConfigController(trialConfigService, trialRepository, webConfig);
+    }
+
+    @Test
+    void createTrialFromJsonFile() {
```
itye-msft, line 48: I would recommend the following naming convention for unit tests:
`<function name>_GiveACondition_ThenExpectedResult`
This may help others to read the test
Alternatively, lets agree on a naming convention for unit test applied by all developers

trial-config-service/src/test/resources/collections/environments/PR.postman_environment.json
```diff
@@ -0,0 +1,29 @@
...
+	"name": "PR",
+	"values": [
+		{
+			"key": "client-id",
+			"value": "901b377a-a4c0-4a72-b390-294cef5c0c74",
+			"enabled": true
+		},
+		{
+			"key": "client-secret",
+			"value": "wq99fHb3Q9Vg_LfsZ_pNEG6~XR_h8um6u4",
```
itye-msft, line 12: Is this a real secret?
If so, it's now compromised. 
There should be a different way to obtain secrets

itye-msft (changes requested):

trial-config-service/.github/workflows/pr-test.yml
```diff
@@ -0,0 +1,39 @@
+name: Test Pull Request
```
tamirkamara, line 1: I don't think this works anymore. You should delete this from the file and add this service name to the workflow in the common folder (/.github/workflows/docker...)

trial-config-service/.github/workflows/sonar.yml
```diff
@@ -0,0 +1,39 @@
+name: Sonar
```
tamirkamara, line 1: Same comment for this workflow.

trial-config-service/pom.xml
```diff
@@ -0,0 +1,160 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">  
+  <modelVersion>4.0.0</modelVersion>  
+  <parent> 
+    <groupId>org.springframework.boot</groupId>  
+    <artifactId>spring-boot-starter-parent</artifactId>  
+    <version>2.3.6.RELEASE</version>
+    <relativePath/>  
+    <!-- lookup parent from repository --> 
+  </parent>  
+  <groupId>uk.ac.ox.ctsu</groupId>
```
tamirkamara, line 12: The updated group should be: **uk.ac.ox.ndph.mts**

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trialconfigservice/TrialConfigServiceApplication.java
```diff
@@ -0,0 +1,15 @@
+package uk.ac.ox.ndph.mts.trialconfigservice;
```
tamirkamara, line 1: on the other services (sample/practitioner) we had a pattern of trial_config_service which is similar to the maven artifact id defined on the pom.
Not sure what is correct, but we should decide on 1 convention

trial-config-service/src/main/resources/application.properties
tamirkamara, line 15: This looks like a real secret / password. I think starting fresh we should commit those and fetch them in other means.

tamirkamara (changes requested):

trial-config-service/src/main/resources/application.properties
shajidissa: should these be best placed in the key vault ?

trial-config-service/src/test/resources/collections/environments/PR.postman_environment.json
shajidissa: @mdp0 any ideas how we can hide this in the postman configuration ?

trial-config-service/.github/workflows/pr-test.yml
shajidissa: I have removed this file but added

      - name: Publish Test Report
          if: ${{ always() }}
          uses: actions/upload-artifact@v2
          with:
            name: postman-reports
            path: test-reports/newman-report.html

to docker-build-push

trial-config-service/.github/workflows/sonar.yml
shajidissa: removed

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trialconfigservice/TrialConfigServiceApplication.java
shajidissa: Renamed to trial_config_service. We will now use this convention

trial-config-service/pom.xml
shajidissa: Renamed to uk.ac.ox.ndph.mts

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trialconfigservice/config/WebConfig.java
shajidissa: moved to constructor

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trialconfigservice/config/WebConfig.java
shajidissa: Yes, absolutely right, changed to camelCase

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trialconfigservice/config/SecurityConfig.java
shajidissa: Added to tech debt

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trialconfigservice/config/WebConfig.java
shajidissa: Moved to statics

trial-config-service/src/test/resources/config.json
```diff
@@ -0,0 +1,15 @@
+{
+  "trial": [
+    {
+      "id": "test_1",
+      "trialName":"Hello World Trial",
+      "trialId":"prefix/1",
```
eladiw, line 6: currently there is a bit of duplication. for eaxmple I am using similar json config in the mts trial deployment repo.
in any case, I think trialId should not have '/' as it might cause issues, e.g. If im using it for resources names...
how about prefix-1

trial-config-service/src/test/resources/config.json
```diff
@@ -0,0 +1,15 @@
+{
+  "trial": [
```
eladiw, line 2: trial or trials?

trial-config-service/src/test/resources/config.json
```diff
@@ -0,0 +1,15 @@
...
+    {
+      "id": "test_1",
+      "trialName":"Hello World Trial",
+      "trialId":"prefix/1",
+      "trialSites":[
```
eladiw, line 7: here it says 'sites' where the parent is singular 'trial'

trial-config-service/src/test/resources/collections/Trial_Config_Service.postman_collection.json
```diff
@@ -0,0 +1,381 @@
+{
+	"info": {
```
eladiw, line 2: how was this file generated? can we add to the readme how to update this

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trial_config_service/controller/TrialConfigController.java
```diff
@@ -0,0 +1,106 @@
...
+        this.webClient = webClient;
+        this.gitRepo = gitRepo;
+    }
+
+    public TrialConfigController(TrialConfigService trialConfigService, String baseUrl) {
+        this.trialConfigService = trialConfigService;
+        this.webClient = WebClient.create(baseUrl);
+    }
+
+    @GetMapping("/trial")
```
eladiw, line 49: why is GET /trial
creates a repo

and POST /trial
also creates a repo (from json)

Seems like the first should also be a post with empty body. wdyt

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trial_config_service/model/Person.java
```diff
@@ -0,0 +1,84 @@
...
+    @Column(unique = true)
+    private String azureOid;
+
+    @JsonIgnore
+    /*
+    This is a dummy relationship - it won't be a direct 1:1 here but will be mediated by Roles.
+    Dummied for the purpose of story 170.
+     */
+    @OneToOne
+    @JoinColumn(name = "trialsite_id")
```
eladiw, line 38:
```suggestion
    @JoinColumn(name = "trial_site_id")
```

?

trial-config-service/src/main/resources/db/changelog/changelog-master.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
eladiw, line 1: how are these files generated?

eladiw (changes requested):
some comments

trial-config-service/src/test/resources/config.json
darrengsmithson: agree - will change

trial-config-service/src/test/resources/config.json
darrengsmithson: not sure - I think we should leave as trial until clarification

trial-config-service/src/test/resources/config.json
darrengsmithson: Cant a Trial have many TrialSites? If so this is correct.

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trial_config_service/controller/TrialConfigController.java
darrengsmithson: Yeah I understand the confusion - I created a POST to accept the JSON data in the requestBody - this is correct as we create a Trial (and child objects) from the Json data. The other endpoint 'GET' is used to read data from a file retrieved from GIT, this also creates a Trial (and child objects). The reason for having the slightly incorrect use of GET is this - 
I wanted to keep the endpoints - trail/trial-config the same for both methods of reading/creating Trials and only have the verb as the difference. Once we get close to finalising the solution and decide whether to use GIT or JSON fed directly into the request we can remove 1 method and ensure that it is a POST. Having the endpoints the same but using different (even tho its slightly incorrect) seemed the best option as the changes to test scripts etc would smaller when changing the verb rather than changing the endpoint when we decide to use only 1 way to create a trial. The other option of changing the endpoint (to ensure we dont get an ambiguous mapping exception) per data retrieval type would also be ugly IMO - e.g trial-config/trial/git, trial-config/trial/json..

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trial_config_service/model/Person.java
darrengsmithson: I think trialsite_id is correct - its the id for a trialSite? We use the following pattern when mapping entites via the id - obj_id - this is used throughout all models - in this case the object is  trialSite.

trial-config-service/src/test/resources/config.json
darrengsmithson: We dont have trialId in the config.json file we are using from GIT so I will remove this trialId ref from this 'test' resource - it was probably copied from the initial config file from a few weeks ago..

trial-config-service/src/main/resources/db/changelog/changelog-master.xml
darrengsmithson: These files are currently edited when we add edit the DB schema. We should start revisioning them and then maybe start using Liquibases generate functionality - that said Im not sure how that works in a container/microservice environement?

trial-config-service/src/test/resources/config.json
eladiw: I mean, if 'trial' is an array then it should be 'trials'
if it is singular, then 'trialSites' should be 'trialSite'.

both singular or plural

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trial_config_service/controller/TrialConfigController.java
eladiw: ok. so lets just add a comment that this should be later changed

eladiw (approved):
approved with minor comments

trial-config-service/src/test/resources/config.json
darrengsmithson: file has been removed as a new testing approach has been agreed.

trial-config-service/src/test/resources/config.json
darrengsmithson: file has been removed as a new testing approach has been agreed.

trial-config-service/src/main/java/uk/ac/ox/ndph/mts/trial_config_service/controller/TrialConfigController.java
darrengsmithson: added a TODO in the code as per Elads suggestion

trial-config-service/src/test/resources/collections/Trial_Config_Service.postman_collection.json
darrengsmithson: file has been removed as a new testing approach has been agreed.

tamirkamara (approved):

PR #11, opened 2020-12-31
Arts 98

darrengsmithson:
Adding Role to the trial config. The changed files in this commit will change again shortly as we remove authentication and retrieve the trial-config json file from git


darrengsmithson:
Conflicts resolved

PR #12, opened 2020-12-31
[ARTS-98]

darrengsmithson:
Creating roles tables in liquibase, adding role functionality when reading json, adding unit tests, updating pom and app props (set flag to true to ensure liquibase changes work after changing changeset)


mdp0 (changes requested):
Please delete comments - unless they're describing a particularly knotty piece of logic or a dangerous trap you can't avoid leaving.

darrengsmithson:
yep will do, was gonna remove them but I left them in as I expect them to be moved into separate Test methods when we get a little closer to the finished article.

shajidissa (approved):
All good

PR #14, opened 2021-01-03
Enable checkstyle for all projects

tamirkamara:
1. Enable checkstyle plugin with a single definition file
2. Fix style issues
3. Update docker files and remove mvnw


pom.xml
```diff
@@ -99,6 +102,38 @@
...
+                <executions>
+                    <execution>
+                        <id>validate</id>
+                        <phase>validate</phase>
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                    </execution>
+                </executions>
+<!--                This doesn't work for some reason so had to reference the project directly. -->
```
itye-msft, line 128: Is there a user story to solve that?

itye-msft (approved):
Approved with one question...

pom.xml
tamirkamara: Currently no. I just kept it here for reference...

tamirkamara: I removed it now.

shajidissa (approved):

PR #16, opened 2021-01-04
Arts 260 testbranch

sameerapurini:
The PR covers the mocha API automation tests using Chai assertions. The reports are generated in the form of HTML reports which can launched on any local browser.

A spike Jira ticket is added to ARTS backlog in order to keep the changes in loop.

https://ndph-arts.atlassian.net/browse/ARTS-260


tamirkamara (commented):
1. This PR contains report files. Are you sure we need to commit those?
2. I'm unsure about having "api-tests" as a root folder, but will take it offline.

PR #20, opened 2021-01-05
Arts 285

darrengsmithson:
Tidying exceptions, ensuring unix + windows file separator is used and force calling destroy as the java feedback via annotations isnt working..Ive tried lots of ways without luck..


shajidissa (approved):

darrengsmithson:
We create a directory to hold the git json file which is then deleted as soon as its read, I initially used the classes/resources directory for this but Windows didnt like it as it didnt parse the path correctly. Ive played around with a few scenarios and ended up just creating a dir at the base of the app called gitRepo - however this is determined by the project set-up in Eclipse/IntelliJ so Im not sure how this will work in Azure. That said we delete the directory in a finally block so it shouldnt matter too much as long as we have permissions to create the directory - which, again Im guessing we should as it should be running in a container? So in a nutshell I think its ok but we may, possibly have an issue in Azure...

mdp0 (approved):

PR #36, opened 2021-01-14
[ARTS-159]testbranch

sameendph:
The PR contains API automation test against trial config service


sameendph:
deleted due to conflicting files

PR #45, opened 2021-01-14
[ARTS-286] Role service

KateSant:
## Description

NB this is the CLEAN branch (avoiding the rewritten git history issues).

New role-service.   Plus refactor of trial-config-service to make it call role-service and associated docker compose.

### Changes
task [ARTS-286] / story [ARTS-98]

[ARTS-286]: https://ndph-arts.atlassian.net/browse/ARTS-286
[ARTS-98]: https://ndph-arts.atlassian.net/browse/ARTS-98


docker-compose.yml
```diff
@@ -23,10 +23,12 @@ services:
     image: ghcr.io/ndph-arts/trial-config-service:${GITHUB_SHA}
     depends_on:
       - sqlserver
+      - role-service
     environment:
       SERVER_PORT: "81"
       JDBC_DRIVER: ${JDBC_DRIVER}
       JDBC_URL: ${JDBC_URL}
+      ROLE_SERVICE: "http://role-service-container:82"
```
mdp0, line 31: Shouldn't this just be the service name, rather than container? I.e. role-service:82

role-service/pom.xml
```diff
@@ -0,0 +1,146 @@
...
+            <version>2.3.1.RELEASE</version>
+        </dependency>
+        <dependency>
+            <groupId>com.microsoft.sqlserver</groupId>
+            <artifactId>mssql-jdbc</artifactId>
+            <version>8.4.1.jre11</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
```
tamirkamara, line 91: This exists in the parent...

tamirkamara (changes requested):

docker-compose.yml
KateSant: Thanks Mark, good point. I have changed this.

role-service/pom.xml
KateSant: Thanks Tamir, good point.  I have cleaned up this pom.

build-tools/src/main/resources/checkstyle-suppressions.xml
```diff
@@ -12,4 +12,8 @@
     <suppress checks="MissingJavadocMethod" files="trial-config-service.*\.java"/>
     <suppress checks="MissingJavadocType" files="trial-config-service.*\.java"/>
 
-</suppressions>
\ No newline at end of file
+    <!-- should we do this across all services? -->
+    <suppress checks="MissingJavadocMethod" files="role-config-service.*\.java"/>
+    <suppress checks="MissingJavadocType" files="role-config-service.*\.java"/>
+
+</suppressions>
```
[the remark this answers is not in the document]
KateSant: Thanks Tamir good point.   I have removed these.  Same for the trial-config-service suppressions I think?

role-service/pom.xml
```diff
@@ -0,0 +1,146 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.modelmapper</groupId>
+            <artifactId>modelmapper</artifactId>
+            <version>2.3.5</version>
+        </dependency>
+
+    </dependencies>
+    <dependencyManagement>
```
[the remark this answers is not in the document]
KateSant: Thanks Tamir, good point.  I have cleaned up this pom.

role-service/src/main/java/uk/ac/ox/ndph/mts/role_service/controller/RoleController.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.web.bind.annotation.RequestBody;
+import uk.ac.ox.ndph.mts.role_service.model.Role;
+import uk.ac.ox.ndph.mts.role_service.model.RoleDTO;
+import uk.ac.ox.ndph.mts.role_service.model.RoleRepository;
+
+import javax.validation.Valid;
+
+
+@RestController
+@RequestMapping("/roles")
```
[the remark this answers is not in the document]
KateSant: I'm not sure - there is a debate on singular vs plural convention for REST endpoints.  We generally use plural, see: https://docs.microsoft.com/en-us/azure/architecture/best-practices/api-design   But I do not have a strong view - happy to change to singular if you want?  We should be consistent across our services though.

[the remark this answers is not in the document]
tamirkamara: You're right. So we'll need to adjust the other services (something for another day)

build-tools/src/main/resources/checkstyle-suppressions.xml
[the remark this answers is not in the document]
tamirkamara: Right. you can delete those...

role-service/src/test/java/uk/ac/ox/ndph/mts/role_service/controller/RoleControllerTest.java
```diff
@@ -0,0 +1,101 @@
...
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(jsonRole)
+                .accept(MediaType.APPLICATION_JSON))
+                .andDo(MockMvcResultHandlers.print())
+                .andExpect(status().isOk());
+
+    }
+
+    @Test
+    void givenInvalidRole_whenPost_thenReturn400()
```
tamirkamara, line 63: Do we need to validate that the role name is unique?

role-service/src/main/java/uk/ac/ox/ndph/mts/role_service/model/Role.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.validation.constraints.NotBlank;
+import java.time.LocalDateTime;
+
+@Entity
+@EntityListeners(AuditingEntityListener.class)
+@Audited
+public class Role {
+    @Id
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "role_sequence")
+    @SequenceGenerator(name = "role_sequence", sequenceName = "role_sequence", allocationSize = 1)
```
tamirkamara, line 29: This means that we can have roles defined with a specific ID as described in the trial config example.  I don't have a specific ask here...

KateSant: Yes.  I wonder whether it makes more sense to reference them by (unique) role name in the json, instead of a generated int?

role-service/src/test/java/uk/ac/ox/ndph/mts/role_service/controller/RoleControllerTest.java
KateSant: I am not sure - maybe the ID is the unique property to rely on?   OR if the name is the true unique ID, then perhaps we don't need the numeric ID?   I am in discussion with Adrian about the various properties of roles and permissions today, so I will raise this with him and I'll deal with this in branch 101 if it's to be unique.  OK to leave as-is for this PR?

tamirkamara (dismissed):

role-service/src/test/java/uk/ac/ox/ndph/mts/role_service/controller/RoleControllerTest.java
KateSant: Change of plan - I have confirmed that the name is unique and we can use it as the ID, so I've added the validation and error handling to pick up duplicates.

trial-config-service/pom.xml
```diff
@@ -111,6 +111,13 @@
       <artifactId>json-path</artifactId>
       <version>2.2.0</version>
     </dependency>
+
```
shajidissa, line 114: Kate can you remove this dependency please

role-service/pom.xml
```diff
@@ -0,0 +1,111 @@
...
+            <version>8.4.1.jre11</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.16</version>
+            <scope>provided</scope>
+        </dependency>
+
```
shajidissa, line 96: Kate can you remove the lombok dependency please

tamirkamara: @shajidissa, @KateSant has another task about permissions on the role-service which is starting right after this. I think we should approve this one and remove lombok as part of the other task. wdyt?

tamirkamara (dismissed):

role-service/pom.xml
KateSant: Why are we dropping lombok?  (It has come into fashion and gone out again in the time this feature has been in dev!)  Yes please I agree with @tamirkamara I think it would be better as a separate task across all services, I can then unblock Sameera to get on with QA on roles.

mdp0 (approved):

tamirkamara (approved):

role-service/pom.xml
KateSant: Discussed in another thread with @shajidissa and agreed.

PR #46, opened 2021-01-15
Docker compose improvements

shajidissa:
improvements to docker compose


docker-compose.yml
```diff
@@ -55,3 +57,16 @@ services:
         retries: 6
     ports:
       - "1433:1433"
+
+  sql-data:
```
marrobi, line 61: is `sql-data` the best name? or `sql-mts`, or `sql-mts-data` maybe?

docker-compose.yml
```diff
@@ -55,3 +57,16 @@ services:
...
+    environment:
+      SA_PASSWORD: ${SAPASSWORD}
+      ACCEPT_EULA: "Y"
+    healthcheck:
+        test: ["CMD", "/opt/mssql-tools/bin/sqlcmd","-S sql-data,1434 -U sa -P ${SAPASSWORD} -Q 'SELECT * FROM INFORMATION_SCHEMA.TABLES'"]
+        interval: 10s
+        timeout: 10s
+        retries: 6
+    ports:
+      - "1434:1433"
```
marrobi, line 72: I believe we only need to specify ports if we want to service to be exposed externally. I'd query if either SQL server need to be exposed externally. If not, try not exposing ports at all, they should both be accessible on 1433...

Doesn't really matter I guess as using compose for dev, but aware the port numbers are getting messier.

marrobi (commented):
Not blockers, things that may want to think about.

docker-compose.yml
shajidissa: changed name to sql-mts-data

docker-compose.yml
shajidissa: QA are connecting to sql on their host machines. Useful for now.

Yeh, the ports will get messier once we get more micro services.

mdp0 (approved):

marrobi (approved):

PR #51, opened 2021-01-18
Fix health checks, add script to check compose services, add compose to workflow

shajidissa:
health check issue


marrobi:
@shajidissa If SQL does not come up before trial-config-service get an issue.

```trial-config-service_1  | org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'liquibase' defined in class path resource [org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration$LiquibaseConfiguration.class]: Invocation of init method failed; nested exception is liquibase.exception.DatabaseException: com.microsoft.sqlserver.jdbc.SQLServerException: Login failed for user 'sa'. ClientConnectionId:ddb3cc01-eec5-40d7-8a5d-a1c2dd86f3d4```

If then run `docker-compose up trail-config-service` the service comes up. Options:

1. Start SQL services first
2. Have java retry for a connection
3. Have an init container on the service that waits
4. User a restart policy `restart: on-failure`

I will try 4 - and if works push changes.

**Comment by marrobi** [comment]
@shajidissa that sorts it - same approach is used in FHIR.

Do you need to get this merge to unblock people or want me to add the compose into CI to ensure all is good on each PR (or can create a new PR after this one if @mdp0  you have a ticket)?

**Comment by shajidissa** [comment]
> @shajidissa that sorts it - same approach is used in FHIR.
> 
> Do you need to get this merge to unblock people or want me to add the compose into CI to ensure all is good on each PR (or can create a new PR after this one if @mdp0 you have a ticket)?

**Review by mdp0 — APPROVED** [review]

File: build-tools/docker/check-docker-compose-services.sh [hunk 1]
```diff
@@ -0,0 +1,21 @@
...
+        echo "$1 is healthy"
+        return 0
+    else
+        echo "Waiting for $1"
+        return 1
+    fi
+}
+
+while ! is_healthy trial-config-service; do sleep 5; done
+while ! is_healthy practitioner-service; do sleep 5; done
```
tamirkamara, line 21: Those would run infinitely, right?
I think we should add some protection here, like max 90 seconds or so...

build-tools/docker/check-docker-compose-services.sh
```diff
@@ -0,0 +1,21 @@
+export SAPASSWORD=$(openssl rand -base64 12)
```
tamirkamara, line 1: build-tools is a java project, and looks like this file is more around ci checks. on the deployment-config repo we added a ".ci" folder with scripts like this one. What about doing the same here?

.github/workflows/docker-build-push.yml
```diff
@@ -41,3 +41,13 @@ jobs:
...
           docker push $GHCR/"${{ matrix.service_name }}":${{ github.sha }}
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Start services using Docker Compose
+        run: |
+          chmod +x ./build-tools/docker/check-docker-compose-services.sh
```
tamirkamara, line 52: Git supports +x for files, so you can save this step by doing this:
https://stackoverflow.com/questions/40978921/how-to-add-chmod-permissions-to-file-in-git/40979016

tamirkamara (changes requested):

.github/workflows/docker-build-push.yml
```diff
@@ -41,3 +41,13 @@ jobs:
         run: |
           docker build --build-arg SVC="${{ matrix.service_name }}" -t $GHCR/"${{ matrix.service_name }}":${{ github.sha }} -f "${{ matrix.service_name }}"/Dockerfile .
           docker push $GHCR/"${{ matrix.service_name }}":${{ github.sha }}
+  test:
```
shajidissa, line 44: @marrobi better name for this, instead of test.

build-tools/docker/check-docker-compose-services.sh
marrobi: Doh. Meant to that in the action job timeout. Will fix.

build-tools/docker/check-docker-compose-services.sh
marrobi: Will move.

.github/workflows/docker-build-push.yml
marrobi: @shajidissa suggest one, I thought "build" followed by "test" was ok!

.github/workflows/docker-build-push.yml
marrobi: Didn't know that. Will give it a go.

build-tools/docker/check-docker-compose-services.sh
shajidissa: moved to .ci directory

.github/workflows/docker-build-push.yml
shajidissa: applied git update-index --chmod=+x .ci/docker/check-docker-compose-services.sh and removed 'chmod' line..

build-tools/docker/check-docker-compose-services.sh
shajidissa: Marcus is this a property on a job ?

timeout-minutes: 1

marrobi: Yes, that's what is meant to do. 1 minute should be plenty I think?

build-tools/docker/check-docker-compose-services.sh
marrobi: Updated the workflow too? It's failing as can't find it.

marrobi:
@shajidissa as the first build stage is a matrix can't run the second stage on the same agent. Hence all images have to get rebuilt - no caching. Think post this PR it's worth considering if is a better way to do this.

marrobi:
@tamirkamara  @mdp0  can you give this another review.

@shajidissa remind me tomorrow to talk about how we can optimise the workflow. The images need to be built twice at present - one each in the matrix and once in the "test" stage.

PR #53, opened 2021-01-18
Test potential compose workflow

marrobi:
## Description
Testing compose workflow

### Changes
- [ARTS-###] - some description

### Checklist:

- [ ] Branch name follows convention (feature/arts-###-short-name OR fix/arts-###-short-name)


marrobi:
@shajidissa @mdp0 @tamirkamara just wanted to have a play as didn't like the fact we were building the same images on different runners multiple times. Might be one to look at and if current setup becomes an issue look at as alternative. Main advantage is images are built once, and also no images are pushed until all tests are completed. However, no parallel runner (but time is similar/lower as no rebuild), and all test results are on a single report.

PR #57, opened 2021-01-19
[ARTS-98] Role Service API Automation Tests

sameerapurini:
The PR includes automation api tests written around role service and trial config service end points. 

Jira under test is https://ndph-arts.atlassian.net/browse/ARTS-98.

### DIFF
--- a/api-tests/data/role-service/roleservice.js
+++ b/api-tests/data/role-service/roleservice.js
@@ -2,6 +2,44 @@ const validRole = {
     "id": makeid(7),
 }
 
+const duplicateRole = {
+    "id": "Sameera",
+}
+
+const emptyString = {
+    "id": "",
+}
+
+const tooLongId = {
+    "id": makeid(256),
+}
+
+const trialConfigPost = {
+
+    "id": "Sameera",
+    "trialName": "Hello World Trial",
+
+    "trialSites": [
+        {
+            "siteName": "(Root node)",
+            "siteType": "CCO"
+        },
+        {
+            "siteName": "(Root node dup)",
+            "siteType": "LLC"
+        }
+    ],
+    "roles": [
+        {
+            "id": "superuser"
+        },
+        {
+            "id": "country-admin"
+        }
+    ]
+}
+
+
 function makeid(length) {
     let result = '';
     const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
@@ -14,4 +52,8 @@ function makeid(length) {
 
 module.exports = {
     validRole,
+    duplicateRole,
+    emptyString,
+    tooLongId,
+    trialConfigPost,
 }
--- a/api-tests/data/role-service/roleservice.js
+++ b/api-tests/data/role-service/roleservice.js
@@ -2,10 +2,6 @@ const validRole = {
     "id": makeid(7),
 }
 
-const duplicateRole = {
-    "id": "Sameera",
-}
-
 const emptyString = {
     "id": "",
 }
@@ -52,7 +48,6 @@ function makeid(length) {
 
 module.exports = {
     validRole,
-    duplicateRole,
     emptyString,
     tooLongId,
     trialConfigPost,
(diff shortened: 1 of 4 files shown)


sameerapurini:
Closing the pull request to re-open once addition of all the tests is completed

KateSant:
Thanks Sameera for helping me get mocha-chai running locally.  For other developers these are the commands:

set BASE_URL=http://localhost:82
mts-services\api-tests>npm run test:awesome

(And you need Node.js installed)

api-tests/specs/role-service/roleservice.js
```diff
@@ -0,0 +1,41 @@
...
+    });
+
+    it('User is shown a bad request error when the role name field exceeds the expected string length', async () => {
+        const response = await baseRequest.post('/roles').send(requests.tooLongId);
+        expect(response.status).to.equal(400);
+        expect(response.text).to.contain("Role ID is too long")
+    });
+
+    it('User is able to view the created roles', async () => {
+        const response = await baseRequest.get('/roles?page=1&size=2');
```
KateSant, line 36: Page index starts at 0

api-tests/specs/role-service/roleservice.js
```diff
@@ -0,0 +1,41 @@
...
+        const response = await baseRequest.post('/roles').send(requests.validRole);
+        expect(response.text).to.contain("id")
+        expect(response.status).to.equal(HttpStatus.OK)
+    });
+
+    it('User is prevented from saving the new role if the name is already used for an existing role', async () => {
+        const roleName = requests.validRole
+        let response = await baseRequest.post('/roles').send(roleName);
+        response = await baseRequest.post('/roles').send(roleName);
+        expect(response.status).to.equal(409)
```
KateSant, line 19: Is it better to use HttpStatus.CONFLICT rather than 409?    It is the same thing of course but might be more understandable for future developers?

api-tests/specs/role-service/roleservice.js
```diff
@@ -0,0 +1,41 @@
...
+        expect(response.text).to.contain("Duplicate role ID")
+    });
+
+    it('User is shown a bad request error when the role name field is left empty', async () => {
+        const response = await baseRequest.post('/roles').send(requests.emptyString);
+        expect(response.status).to.equal(400);
```
KateSant, line 25: As above, is there something like HttpStatus.BADREQUEST?

api-tests/data/role-service/roleservice.js
```diff
@@ -0,0 +1,28 @@
...
+const emptyString = {
+    "id": "",
+}
+
+const tooLongId = {
+    "id": makeid(256),
+}
+
+
+function makeid(length) {
```
KateSant, line 14: It looks as though the purpose of this function is to generate a random alphanumeric string?    I expect you will want to reuse this in other tests going forward, would it be worth naming it something more generic like "makeRandomString"?   
I don't know much about javascript libraries but I wondered whether there is a library function for generating strings, to save you writing this loop by hand?

api-tests/specs/role-service/roleservice.js
sameerapurini: that's a very valid point, but the response status is sent back as either 404 or 409 but no error text attached to the status as below
I can use an additional assertion like,  expect(response.body).to.contain(Conflict) if that helps

`{"timestamp":"2021-01-21T11:03:54.070+00:00","status":409,"error":"Conflict","message":"Duplicate role ID","path":"/roles"}`

sameerapurini: Thanks Kate, I have found a way to handle it :)

api-tests/data/role-service/roleservice.js
sameerapurini: I have updated the name as suggested but with regards to using a library and avoiding the loop, its not possible in this scenario as I am pulling the length of the string as well for 2 scenarios.

KateSant (approved):

sameendph (approved):
Approving

PR #58, opened 2021-01-20
Fixed checkstyle setup to allow checkstyle:check Maven goal to run against individual project modules

archiemaskill:
A problem with the parent POM setup is that switching to a module directory (e.g. trial-config-service) and running 'mvn install' (or mvn checkstyle:check) results in a CheckStyle error, because the checkstyle.xml file cannot be located.

Details of my fix:

- Deleted the existing suppressions from checkstyle-suppressions.xml because they are not currently suppressing anything (the rules they were trying to suppress were removed from checkstyle.xml, and so suppression is unnecessary).
- Removed the SuppressionFilter from checkstyle.xml, because there is an alternative & tidier way to do this in pom.xml
- In the parent pom.xml, build-tools is now a declared module of the project, so it will be built when the project is built. It's also a declared dependency of the maven-checkstyle-plugin, so it will be built before the plugin is run. This also allows the plugin to access the checkstyle XML files from within the build-tools artifact in the Maven repository.
- It was necessary to change 'mvn package' to 'mvn install' in each Dockerfile, so that build-tools would be installed in each image's local Maven repository. This allows each service project to refer to the checkstyle files. Without this, the build would break when running 'docker-compose up'.

I've verified that:
  * the project can still be built from the parent POM by running 'mvn install'
  * the services can be built individually by cd-ing to their directory and running 'mvn install'
  * the suppressions file works when there is something in there
  * bad code will still halt the build
  * 'docker-compose up' produces a running system

There is no linked ARTS story for this change.


archiemaskill:
Looks to me like GitHub Actions' Docker build and push is failing because it runs 'mvn verify' which is one lifecycle phase short of what's needed. It needs to install build-tools to its local Maven repo (by running 'mvn install') for this method to work.

archiemaskill:
Ignore this pull request for the time being; Mark and I think we've found a better way to achieve the same thing, but are waiting for another story to complete first.

archiemaskill:
Closing for now. If we decide the work is still useful later on, will re-open.

PR #62, opened 2021-01-21
[ARTS-101] Permissions

KateSant:
## Description
Extend role-service to deal with permissions.  

New endpoints for creating role-permission mappings.

Also removes lombok dependency from role-service.

### Changes
[ARTS-101]
### Checklist:

- [x] Branch name follows convention (feature/arts-###-short-name OR fix/arts-###-short-name)

[ARTS-101]: https://ndph-arts.atlassian.net/browse/ARTS-101


role-service/src/main/java/uk/ac/ox/ndph/mts/role_service/controller/dtos/PermissionDTO.java
```diff
@@ -0,0 +1,19 @@
+package uk.ac.ox.ndph.mts.role_service.controller.dtos;
+
+
+import javax.validation.constraints.NotBlank;
+
+
+public class PermissionDTO {
+
+    @NotBlank (message = "Permission ID cannot be empty")
+    private String id;
```
tamirkamara, line 10: Should this also include a max notation like you did in the Role one?

role-service/src/main/java/uk/ac/ox/ndph/mts/role_service/service/RoleService.java
```diff
@@ -1,26 +1,55 @@
...
+        Optional<Role> roleOptional = roleRepository.findById(roleId);
+        if (roleOptional.isEmpty()) {
+            throw new ResponseStatusException(HttpStatus.NOT_FOUND, "Role \"" + roleId + "\" not found.");
+        }
+
+        for (Permission newPermission : newPermissions) {
+            if (!permissionRepository.existsById(newPermission.getId())) {
+                throw new ResponseStatusException(
+                        HttpStatus.BAD_REQUEST,
+                        "Cannot update role \"" + roleId + " because permission \" to permission \" is unknown.");
```
tamirkamara, line 47: Can you convert this to us a string format?

role-service/src/test/java/uk/ac/ox/ndph/mts/role_service/controller/PermissionControllerTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Test
+    void whenGetPaged_thenReceiveSuccess()
+            throws Exception {
+        Permission dummyPermission = new Permission();
+        String dummyId="dummy-permission";
+        dummyPermission.setId(dummyId);
+
+        when(permissionRepo.findAll(PageRequest.of(0, 10))).thenReturn(new PageImpl(Collections.singletonList(dummyPermission)));
+        mvc.perform(get("/permissions?page=0&size=10"))
```
tamirkamara, line 41: In another task I've used MvcUriComponentsBuilder to get the uri of the method so that it won't be listed here and in the tests. What do you think about something like:
`    private String getRoleAssignmentUri() {
        return MvcUriComponentsBuilder.fromMethodCall(on(PractitionerController.class)
                .saveRoleAssignment("987", null)).build().toUriString();
    }`

tamirkamara (changes requested):
Nice, I had a couple of small asks/questions

role-service/src/main/java/uk/ac/ox/ndph/mts/role_service/service/RoleService.java
```diff
@@ -1,26 +1,55 @@
...
         }
+        return roleRepository.save(role);
+    }
+
+    public Role updatePermissionsForRole(String roleId, List<Permission> newPermissions)
+            throws ResponseStatusException {
+
+        Optional<Role> roleOptional = roleRepository.findById(roleId);
+        if (roleOptional.isEmpty()) {
+            throw new ResponseStatusException(HttpStatus.NOT_FOUND, "Role \"" + roleId + "\" not found.");
```
tamirkamara, line 40: In other services we used enums + string format for those. 
An example: https://github.com/NDPH-ARTS/mts-services/blob/main/practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/repository/FhirRepo.java

role-service/src/main/java/uk/ac/ox/ndph/mts/role_service/service/RoleService.java
tamirkamara, line 47: same comment about the string

tamirkamara (changes requested):

role-service/src/main/java/uk/ac/ox/ndph/mts/role_service/service/RoleService.java
KateSant: Thanks @tamirkamara good point.  I've refactored to an enum.

role-service/src/main/java/uk/ac/ox/ndph/mts/role_service/service/RoleService.java
KateSant: Thanks good point - refactored.

shajidissa (dismissed):

role-service/src/main/java/uk/ac/ox/ndph/mts/role_service/controller/dtos/PermissionDTO.java
KateSant: Thanks @tamirkamara .  My initial thinking was that since there is no create-permission endpoint (and the liquibase load won't go through this filter)  then there was no point in any validation here on these IDs -  other than a not-empty check to ensure downstream code can lookup by ID.   However this is probably over-thinking it, and at some point we may extend this to enable permission creation.  I've added the validation in to match the DB field.

sameerapurini:
I have checked the unit test coverage which looks satisfactory, will check the scenario level tests once its merged into master.

sameerapurini (approved):

role-service/src/test/java/uk/ac/ox/ndph/mts/role_service/controller/PermissionControllerTest.java
KateSant: I am not sure.   It adds a lot of complexity and is opaque.   I see the benefit of dynamically discovering the URI, in case it changes.   But is this benefit worth the cost - KISS / self documenting etc?   Not a strong resistance - happy to switch to this if you think it's best - we should standardise one way or the other across all services?

tamirkamara: We decided to go with variables in the tests to make it tidy &simple

tamirkamara (approved):
Nice job!

role-service/src/test/java/uk/ac/ox/ndph/mts/role_service/controller/PermissionControllerTest.java
KateSant: We discussed and agreed to use constants for uris in tests instead.  I have updated RoleControllerTest as well as PermissionControllerTest.

sameerapurini (approved):

PR #79, opened 2021-01-26
[ARTS-473] Tidy practitioner test output

archiemaskill:
Tidied up the outputs detailed in the ARTS-473 story.

Ran a Postman test of creating a practitioner when the FHIR service is down: the response is 502 and message saying "Failed to save practitioner", as expected. Checked the practitioner service logs to ensure sufficient details are still logged.


practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/repository/HapiFhirRepository.java
```diff
@@ -21,7 +20,7 @@
     private static final String PRACTITIONER_ENTITY_NAME = "Practitioner";
 
     private final FhirContextWrapper fhirContextWrapper;
-    private final Logger logger = LoggerFactory.getLogger(HapiFhirRepository.class);
+    private Logger logger = LoggerFactory.getLogger(HapiFhirRepository.class);
```
shajidissa, line 23: Archie, shouldn't this remain

private static final Logger LOGGER

archiemaskill: Not quite sure about the "static" thing really - I think I saw weird threading issues once with a logger, due to a static instance, so I've avoided static loggers ever since (at this point, that might now be me being superstitious). However, the slf4j FAQ mentions they no longer recommend one approach over the other (http://slf4j.org/faq.html#declared_static). The only one of the advantages that I found interesting was that non-static is more IOC-friendly. Which brings me on to the next bit....

I'm a huge fan of "final", but I dropped it in this case to allow me to add a setLogger method. This seemed to me to be the easiest way to allow dependency injection for the purposes of a test - I use it to set a silent mock logger, so that the test ouput isn't polluted with a distracting stacktrace. I wonder if there's a better way though...

archiemaskill: I've now removed setLogger and changed this back to final. Sonar wasn't happy about the log-and-throw antipattern it was seeing so I removed the log and put its message into the RestException that is thrown. This is then logged by the global RestException handler.

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/repository/FhirContextWrapper.java
```diff
@@ -46,12 +46,12 @@ public String prettyPrint(IBaseResource resource) {
...
      */
-    public Bundle executeTransaction(String uri, Bundle input) throws FhirServerResponseException {
+    public Bundle executeTransaction(Bundle input) throws FhirServerResponseException {
         try {
-            return fhirContext.newRestfulGenericClient(uri).transaction()
+            return fhirContext.newRestfulGenericClient(fhirUri).transaction()
                     .withBundle(input).execute();
         } catch (BaseServerResponseException ex) {
-            throw new FhirServerResponseException("Problem executing transaction with bundle at: " + uri, ex);
+            throw new FhirServerResponseException("Problem executing transaction with bundle at: " + fhirUri, ex);
```
tamirkamara, line 54: Can you add this exception text to the enum class in the same package and use is from there?

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/repository/HapiFhirRepository.java
```diff
@@ -64,9 +66,8 @@ public String savePractitionerRole(PractitionerRole practitionerRole) {
         Bundle responseBundle;
         try {
             responseBundle = fhirContextWrapper.executeTransaction(bundle(practitionerRole));
-        } catch (BaseServerResponseException e) {
-            logger.warn(FhirRepo.UPDATE_ERROR.message(), e);
```
tamirkamara, line 72: Is this enum still in use? If not, please remove it.

practitioner-service/src/test/java/uk/ac/ox/ndph/mts/practitioner_service/repository/HapiFhirRepositoryTests.java
```diff
@@ -35,19 +35,26 @@
     @Mock
     Logger SILENT_LOGGER;
```
tamirkamara, line 36: do you still need this?

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/repository/HapiFhirRepository.java
```diff
@@ -43,9 +42,8 @@ public String savePractitioner(Practitioner practitioner) {
         Bundle responseBundle;
         try {
             responseBundle = fhirContextWrapper.executeTransaction(bundle(practitioner));
-        } catch (BaseServerResponseException e) {
-            logger.warn(FhirRepo.UPDATE_ERROR.message(), e);
-            throw new RestException(e.getMessage(), e);
+        } catch (FhirServerResponseException e) {
+            throw new RestException("Failed to save practitioner", e);
```
tamirkamara, line 46: Can you move this to an enum?

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/repository/HapiFhirRepository.java
```diff
@@ -68,9 +66,8 @@ public String savePractitionerRole(PractitionerRole practitionerRole) {
         Bundle responseBundle;
         try {
             responseBundle = fhirContextWrapper.executeTransaction(bundle(practitionerRole));
-        } catch (BaseServerResponseException e) {
-            logger.warn(FhirRepo.UPDATE_ERROR.message(), e);
-            throw new RestException(e.getMessage(), e);
+        } catch (FhirServerResponseException e) {
+            throw new RestException("Failed to save practitioner role", e);
```
tamirkamara, line 70: Can you move this to an enum?

practitioner-service/src/test/java/uk/ac/ox/ndph/mts/practitioner_service/repository/FhirContextWrapperIntegrationTest.java
```diff
@@ -0,0 +1,23 @@
+package uk.ac.ox.ndph.mts.practitioner_service.repository;
+
+import org.hl7.fhir.r4.model.Bundle;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class FhirContextWrapperIntegrationTest {
```
tamirkamara, line 9: 1. What makes this an integration test (btw, I don't think that PractitionerIntegrationTest are it either)?
2. Would something like this work?
```java
@SpringBootTest(properties = { "fhir.uri=something" })
```

tamirkamara (changes requested):

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/repository/FhirContextWrapper.java
archiemaskill: Fixed.

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/repository/HapiFhirRepository.java
archiemaskill: Enum removed.

practitioner-service/src/test/java/uk/ac/ox/ndph/mts/practitioner_service/repository/HapiFhirRepositoryTests.java
archiemaskill: Good catch - have removed.

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/repository/HapiFhirRepository.java
archiemaskill: Moved to enum.

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/repository/HapiFhirRepository.java
archiemaskill: Moved to enum.

practitioner-service/src/test/java/uk/ac/ox/ndph/mts/practitioner_service/repository/FhirContextWrapperIntegrationTest.java
archiemaskill: Ah yes... I thought this was going to be an integration test, but turns out it was half-baked! I've renamed the class, and finished implementing the test.

I thought the same thing when I saw PractitionerIntegrationTests.java, but after pondering for a while I think the original author was testing the boundary between the code we control and the code we don't (i.e. the Spring framework). They also seem to take longer than the other unit tests, so could be they were trying to highlight that aspect too...

tamirkamara (approved):

mdp0 (approved):

PR #103, opened 2021-02-03
[ARTS-404] Get sites from site service

alexbaxter:
## Description

* Added get-all-sites endpoint to site-service `GET ${site-service}/sites`
* Added new endpoint (translating from the FHIR API), matching tests, updated README to document API.
* Site model now has site ID (the FHIR ID), required for matching parentSiteId

### Changes
- [ARTS-404] - get sites from site service

### Checklist:

- [X] Branch name follows convention (feature/arts-###-short-name OR fix/arts-###-short-name)
- [X] I have included a short-meaningful title, description and changes for this PR

[ARTS-404]: https://ndph-arts.atlassian.net/browse/ARTS-404

### DIFF
--- a/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/converter/OrganizationConverter.java
+++ b/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/converter/OrganizationConverter.java
@@ -6,6 +6,8 @@ import org.hl7.fhir.r4.model.Reference;
 import org.springframework.stereotype.Component;
 import uk.ac.ox.ndph.mts.site_service.model.Site;
 
+import static uk.ac.ox.ndph.mts.site_service.converter.SiteConverter.ORG_PREFIX;
+
 /**
  */
 @Component
@@ -34,7 +36,7 @@ public class OrganizationConverter implements EntityConverter<Site, org.hl7.fhir
             // TODO: Check if the Parent Organization already exists.
 
             // Set the Parent Organization if and only if it exists.
-            fhirOrganization.setPartOf(new Reference("Organization/" + site.getParentSiteId()));
+            fhirOrganization.setPartOf(new Reference(ORG_PREFIX + site.getParentSiteId()));
         }
     }
 }
--- a/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/converter/OrganizationConverter.java
+++ b/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/converter/OrganizationConverter.java
@@ -7,7 +7,6 @@ import org.springframework.stereotype.Component;
 import uk.ac.ox.ndph.mts.site_service.model.Site;
 
 /**
- * Implement an EntityConverter for Site
  */
 @Component
 public class OrganizationConverter implements EntityConverter<Site, org.hl7.fhir.r4.model.Organization> {
--- a/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/converter/OrganizationConverter.java
+++ b/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/converter/OrganizationConverter.java
@@ -6,8 +6,6 @@ import org.hl7.fhir.r4.model.Reference;
 import org.springframework.stereotype.Component;
 import uk.ac.ox.ndph.mts.site_service.model.Site;
 
-import static uk.ac.ox.ndph.mts.site_service.converter.SiteConverter.ORG_PREFIX;
-
 /**
  */
 @Component
@@ -36,7 +34,8 @@ public class OrganizationConverter implements EntityConverter<Site, org.hl7.fhir
             // TODO: Check if the Parent Organization already exists.
 
             // Set the Parent Organization if and only if it exists.
-            fhirOrganization.setPartOf(new Reference(ORG_PREFIX + site.getParentSiteId()));
+            fhirOrganization.setPartOf(new Reference("Organization/" + site.getParentSiteId()));
         }
     }
+
 }
--- a/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/converter/OrganizationConverter.java
+++ b/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/converter/OrganizationConverter.java
@@ -6,8 +6,6 @@ import org.hl7.fhir.r4.model.Reference;
 import org.springframework.stereotype.Component;
 import uk.ac.ox.ndph.mts.site_service.model.Site;
 
-/**
- */
 @Component
 public class OrganizationConverter implements EntityConverter<Site, org.hl7.fhir.r4.model.Organization> {
 
--- a/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/model/Site.java
+++ b/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/model/Site.java
@@ -1,5 +1,7 @@
 package uk.ac.ox.ndph.mts.site_service.model;
 
+import java.util.Objects;
+
 /**
  * Site Model
  */
@@ -24,7 +26,7 @@ public class Site {
     }
 
     /**
-     * Site Constructor with two parameters
+     * Site Constructor with name alias and parent
      *
      * @param name the Site name
      * @param alias the Site alias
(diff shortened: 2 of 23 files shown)


alexbaxter:
I can't seem to get test coverage on FhirContextWrapper because it is closely coupled to the actual FHIR API - any ideas?

alexbaxter:
I'm moving this one back to draft because the service depends on some FHIR internals, now I can test them using a local instance.

alexbaxter:
Ready for review now - note I've added the swagger-ui in just for testing (based on the roles-service), can be removed if requested.

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/converter/OrganizationConverter.java
```diff
@@ -7,7 +7,6 @@
 import uk.ac.ox.ndph.mts.site_service.model.Site;
 
 /**
- * Implement an EntityConverter for Site
  */
```
tamirkamara, line 10: [nit] If you deleted the comment text, can you also remove the actual comment pre/suffix?

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/repository/EntityStore.java
```diff
@@ -11,4 +13,6 @@
      * @return the entity stored id.
      */
     String saveEntity(T entity);
+
+    List<T> geAllEntities();
```
tamirkamara, line 17: 1. Typo in the method name.
2. Some of us had a discussion on the naming conventions of methods in the repository layer. Get has a "client" sound to it, but since a repository is similar to a db maybe call it something like "fetch". A question in our internal channel might be good too...

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/service/SiteServiceImpl.java
```diff
@@ -56,4 +59,19 @@ public String save(Site site) {
...
+    /**
+     * Get complete sites list
+     * @return list of sites, never empty
+     * @throws InvariantException if the list from the store is empty
+     */
+    @GetMapping
+    public List<Site> getSites() {
+        final List<Site> sites = this.siteStore.geAllEntities();
+        if (sites.isEmpty()) {
+            throw new InvariantException(Services.NO_ROOT_SITE.message());
```
tamirkamara, line 72: 1. Should we really return an error here? Is it because it's not a normal situation not to have any site?
2. Why specifically this error?
3. I think it's worth to add comments to describe the "why" in the code since it's probably not regular rest behavior.

site-service/src/test/java/uk/ac/ox/ndph/mts/site_service/SiteServiceImplIntegrationTests.java
```diff
@@ -85,4 +88,31 @@ void TestPostSite_WhenValidParentInput_Returns201AndId() throws Exception {
...
+    void TestGetSites_ReturnsList() throws Exception {
+        // Arrange
+        final var org = new Organization()
+                .setName("CCO")
+                .addAlias("Root");
+        org.setId("this-is-my-id");
+        when(repository.getOrganizations()).thenReturn(List.of(org));
+        // Act + Assert
+        this.mockMvc
+                .perform(get("/sites").contentType(MediaType.APPLICATION_JSON))
```
tamirkamara, line 102: Can you move the route string to be a const on the class level and reuse through?

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/repository/FhirContextWrapper.java
```diff
@@ -50,8 +51,25 @@ public Bundle executeTrasaction(String uri, Bundle input) {
...
+     */
+    public <T extends IBaseResource> List<T> toListOfResourcesOfType(Bundle bundle, Class<T> resourceClass) {
+        return new ArrayList<>(BundleUtil.toListOfResourcesOfType(this.fhirContext, bundle, resourceClass));
+    }
+
+    /**
+     * Return a search instance which can be configured and executed, returning a single type of resource
+     * @param uri FHIR endpoint URI
+     */
+    public IQuery<Bundle> search(final String uri, final String resourceName) {
```
tamirkamara, line 68: Is it possible to also pass in the class type rather than its string name?
(you have a nice example in the method above)

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/repository/HapiFhirRepository.java
```diff
@@ -32,18 +34,34 @@
...
+     * Return the list of all organizations. Note this may include organizations that are
+     * not {uk.ac.ox.ndph.mts.site_service.model.Site}s - caller must filter.
+     * @return all organization instances in the store, might be empty, not null
+     */
+    public Collection<Organization> getOrganizations() {
+        try {
+            // TODO: filter organizations by the extension element that identifies them as sites
+            final Bundle responseBundle = fhirContextWrapper
+                    .search(fhirUri, ORGANIZATION_ENTITY_NAME)
+                    .execute();
```
tamirkamara, line 47: Have you chosen to execute at this class and not of the context on purpose?

tamirkamara (changes requested):
Looks nice, I've added a few comments.

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/model/Site.java
```diff
@@ -37,8 +39,26 @@ public Site(String name, String alias, String parentSiteId) {
...
+     * @param alias the Site alias
+     * @param parentSiteId the Site parentSiteId
+     *
+     */
+    public static Site withIdNameAliasAndParent(final String siteId, String name, String alias, String parentSiteId) {
+        final Site site = new Site(name, alias, parentSiteId);
+        Objects.requireNonNull(siteId);
+        site.setSiteId(siteId);
+        return site;
+    }
```
tamirkamara, line 57: Currently we skip validating all kind of rules/checks for model since we assume they are pure with no logic. Also, if we want to include a factory probably best to name it as such and also have it own class?
Not sure about this one... but happy to be persuaded.

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/model/Site.java
```diff
@@ -37,8 +39,26 @@ public Site(String name, String alias, String parentSiteId) {
...
+
```
tamirkamara, line 58: Currently we skip validating all kind of rules/checks for model since we assume they are pure with no logic.
If we want to include a factory probably best to name it accordingly and also have it in its own class?
Not sure about this one... but happy to be persuaded.

alexbaxter: It is in the way of a 'named constructor', so callers won't easily mix up the parent ID and the site ID. I can take out the null-check on site ID, or just replace it with a constructor, if you like.

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/repository/HapiFhirRepository.java
alexbaxter: Yes, since the repository might want to do various searches and already depends on the FHIR API (deals with Bundle, Resources etc.). So this avoids repeating all search methods on the HapiFhirRepository in the FhirContextWrapper.  Given that none of the methods in FhirContextWrapper can be tested effectively, if we add lots of code in the wrapper (that just replicates methods in HapiFhirRepository) we run the risk of getting spuriously bad code coverage. It's just as easy to mock the result of the search() method as it is to mock the whole wrapper, so you can still test HapiFhirRepository.

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/converter/OrganizationConverter.java
alexbaxter: Done

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/repository/EntityStore.java
alexbaxter: Changed to findAll() to match spring-data usage (and findSites() in the SiteService).

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/service/SiteServiceImpl.java
alexbaxter: I went back and forth on this. My thinking was this is the site service so it is supposed to understand how sites work; if there is no root node (i.e. sites tree is empty), then the trial has not been initialized properly, so we should fail fast. Why this error? Having a root site should be an invariant for trials (after initialization), hence the exception name. I'll add a comment.

site-service/src/test/java/uk/ac/ox/ndph/mts/site_service/SiteServiceImplIntegrationTests.java
alexbaxter: Will do

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/repository/FhirContextWrapper.java
alexbaxter: Or maybe a class type instead of a string, since the class name (without package, i.e. the short-name) **is** the resource name.

tamirkamara (approved):
LGTM

tamirkamara (approved):

mdp0 (approved):

PR #114, opened 2021-02-08
[ARTS-599] Common FHIR context/client

tamirkamara:
## Description
Take the core functionality services use and package as a common module to share across the project.
The wrapper includes a few base generic functions that look like could fit all functionality currently implemented in site and practitioner services.

### Changes
- [ARTS-599] common fhir client

### Checklist:

[ARTS-599]: https://ndph-arts.atlassian.net/browse/ARTS-599

### DIFF
--- a/fhir-client/src/main/java/uk/ac/ox/ndph/mts/fhir_client/FhirContextWrapper.java
+++ b/fhir-client/src/main/java/uk/ac/ox/ndph/mts/fhir_client/FhirContextWrapper.java
@@ -1,0 +1,152 @@
+package uk.ac.ox.ndph.mts.fhir_client;
+
+import ca.uhn.fhir.context.FhirContext;
+import ca.uhn.fhir.rest.client.api.IGenericClient;
+import ca.uhn.fhir.rest.gclient.ICriterion;
+import ca.uhn.fhir.rest.gclient.IQuery;
+import ca.uhn.fhir.rest.server.exceptions.BaseServerResponseException;
+import ca.uhn.fhir.rest.server.exceptions.ResourceNotFoundException;
+import ca.uhn.fhir.util.BundleUtil;
+import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.hl7.fhir.r4.model.Bundle;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+import org.springframework.util.StringUtils;
+
+import java.util.List;
+
+/**
+ * Wrapper around a Fhir Context
+ */
+@Component
+public class FhirContextWrapper {
+    private final FhirContext fhirContext;
+    private IGenericClient client;
+
+    @SuppressWarnings("FieldMayBeFinal") // Value injection shouldn't be final
+    @Value("${fhir.uri}")
+    private String fhirUri = "";
+
+    public FhirContextWrapper() {
+        fhirContext = FhirContext.forR4();
+    }
+
+    public FhirContextWrapper(FhirContext fhirContext) {
+        this.fhirContext = fhirContext;
+    }
+
+    /**
+     * Pretty print a resource
+     *
+     * @param resource the resource to print
+     * @return the pretty print String of the resource
+     */
+    public String prettyPrint(IBaseResource resource) {
+        return fhirContext.newJsonParser().setPrettyPrint(true).encodeResourceToString(resource);
+    }
+
+    /**
+     * Execute a bundle transaction to a FHIR endpoint
+     *
+     * @param input the Bundle to save
+     * @return the returned Bundle object from FHIR endpoint
+     */
+    public Bundle executeTransaction(Bundle input) throws FhirServerResponseException {
+        try {
+            return getClient().transaction()
+                    .withBundle(input).execute();
+        } catch (BaseServerResponseException ex) {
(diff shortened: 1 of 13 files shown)


fhir-client/src/main/java/uk/ac/ox/ndph/mts/fhir_client/FhirContextWrapper.java
```diff
@@ -0,0 +1,152 @@
...
+
+        return resourceClass.cast(baseResource);
+    }
+
+    private IGenericClient getClient() {
+        if (this.client == null) {
+            client = fhirContext.newRestfulGenericClient(requireNonBlank(fhirUri, "fhirUri"));
+        }
+        return client;
+    }
```
alexbaxter, line 142: Lazy init of member `client`, means this class is not threadsafe - should be documented at least. Not sure of when instances of FhirContextWrapper are created because that will be done by Spring. Is the IGenericClient threadsafe anyway (docs are unclear)? Is the FhirContext threadsafe (docs say yes and should be reused)?

alexbaxter (commented):
PR still in draft so not approving

fhir-client/src/main/java/uk/ac/ox/ndph/mts/fhir_client/FhirContextWrapper.java
tamirkamara: Current implementations we have create their own IGenericClient on each method. We can you with that option again...
The doc is here: https://hapifhir.io/hapi-fhir/docs/client/generic_client.html
It says both the context and the client are thread safe and can be reused.

Regarding how spring is doing this - I believe this is a singleton, so once created it remains a single instance for the during of the application.

fhir-client/src/main/java/uk/ac/ox/ndph/mts/fhir_client/FhirContextWrapper.java
```diff
@@ -0,0 +1,152 @@
...
+public class FhirContextWrapper {
+    private final FhirContext fhirContext;
+    private IGenericClient client;
+
+    @SuppressWarnings("FieldMayBeFinal") // Value injection shouldn't be final
+    @Value("${fhir.uri}")
+    private String fhirUri = "";
+
+    public FhirContextWrapper() {
+        fhirContext = FhirContext.forR4();
```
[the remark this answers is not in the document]
tamirkamara: Why? 
It looks less explicit (I had to look at it for a few seconds and understand what it means)...
Anyway, I don't have a strong opinion.

fhir-client/src/main/java/uk/ac/ox/ndph/mts/fhir_client/FhirContextWrapper.java
tamirkamara: Another option would be to put this in the constructor...

alexbaxter: Even if the client is threadsafe, the lazy-initialization is not (because it will presumably be done in a request thread) - so creating the client should be moved to an initializer or the constructor (so will be done only once, and threadsafely, since an object can only be created in one thread). The docs say that creation of the client is cheap, so there is no gain in delaying initialization to first use.

fhir-client/src/main/java/uk/ac/ox/ndph/mts/fhir_client/FhirContextWrapper.java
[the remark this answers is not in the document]
alexbaxter: Sorry I don't have a strong grasp of this GitHub UI yet which probably didn't help. There isn't any logic in the constructors at the moment, but if there was you would want to make sure it was always executed. So the safe thing is to have one constructor that does the work, and the others that call it. Using `this(FhirContext.forR4())` from the default no-arg constructor means that the one-arg constructor is always invoked - so we can be sure any logic that is added will always be invoked.

fhir-client/src/main/java/uk/ac/ox/ndph/mts/fhir_client/FhirContextWrapper.java
shajidissa: @tamirkamara realise this is a work in progress, just a reminder, we'll need to cope with timeouts, when a IGenericClient fails to be initialised.

For the FHIR spike, I looked at 

https://github.com/NDPH-ARTS/fhir-service/blob/main/src/main/java/uk/ac/ox/ctsu/arts/fhirservice/helpers/FHIRClientHelper.java

tamirkamara: @alexbaxter I've update the constructor to also include the client init. I removed the other overload since that was used for unit tests and we don't plan/can have a real unit text here anyway.

tamirkamara: @shajidissa what do you mean in cope? Do you mean retry?

alexbaxter (approved):
Code looks OK so I'm approving this, POM changes probably need review by someone else (changes look consistent, at least).

shajidissa (approved):

PR #115, opened 2021-02-08
[Arts 164] assign roles to bootstrap user 

KateSant:
## Description

This PR is an extension of init-service.    

See story https://ndph-arts.atlassian.net/browse/ARTS-164

init-service has created the site, roles, and practitioner.  This story ties them together by making further calls to practitioner-service to assign role X to practitioner Y at site Z.

New roles list under the bootstrap user json config entry (see https://github.com/NDPH-ARTS/global-trial-config/blob/main/config-init-bootstrap-roles.json) 

Some refactoring of init-service to make the IDs of the entities created available to subsequent calls.

Assume that the roles should be assigned to the first site in the site list in the json config.

### Changes
- [ARTS-164] 

### Checklist:

[ARTS-164]: https://ndph-arts.atlassian.net/browse/ARTS-164


init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/service/RoleServiceInvoker.java
```diff
@@ -16,32 +13,22 @@
...
-    }
-    public RoleServiceInvoker(WebClient webClient) {
-        this.webClient = webClient;
-    }
+    public RoleServiceInvoker() { }
+    public RoleServiceInvoker(WebClient webClient) { super(webClient); }
 
     @Override
-    public Role send(Entity role) throws DependentServiceException {
+    protected String create(Entity role) throws DependentServiceException {
```
darrengsmithson, line 26: Could we move the try catch for the DependentServiceException to the abstract class - saves all sub classes writing the same code?

init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/service/RoleServiceInvoker.java
```diff
@@ -16,32 +13,22 @@
...
 
+        String uri = roleService + "/roles";
```
darrengsmithson, line 28: maybe create endpoint api refs in the properties?

init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/service/SiteServiceInvoker.java
```diff
@@ -3,59 +3,51 @@
...
+            return responseFromSiteService.getId();
+
         } catch (Exception e) {
             LOGGER.info("FAILURE siteService {}", e.getMessage());
             throw new DependentServiceException("Error connecting to site service");
         }
     }
 
-    public void execute(List<Site> sites) throws NullEntityException {
+    public List<String> execute(List<Site> sites) throws NullEntityException {
```
darrengsmithson, line 40: Could we create a generic execute method using <Entity> as this code is duplicated in other invokers?

darrengsmithson (changes requested):

init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/service/RoleServiceInvoker.java
KateSant: Thanks Darren, I've made this change.

init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/service/RoleServiceInvoker.java
KateSant: Thanks Darren, I've made this change. Had to use a wildcard in the props file for this one /practitioner/%s/roles but spring property injection seems to handle that ok.

init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/service/SiteServiceInvoker.java
KateSant: Thanks Darren.   I've had a go at this but not sure whether it improves code clarity or not, on balance.   The execute methods were all a little different (the one in SiteServiceInvoker returns the site IDs to the caller, the one in RoleServiceInvoker returns void since those IDs aren't needed, and the one in PractitionerServiceInvoker does a bit more work with a second rest call).   I've had a go at extracting a method which will work for the site and role cases, leaving the practitioner case on its own.  That leaves the method both overloaded and overridden, which is a bit trixy for the future reader.  wdyt?

darrengsmithson: yeah thats the approach I took when I played with it before adding the comment - you could change List<String> siteIDs = new ArrayList<>(); to List<String> entityIDs = new ArrayList<>(); or just List<String> IDs = new ArrayList<>();

KateSant: Good point that's the wrong name - fixed.

darrengsmithson (dismissed):

init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/model/IDResponse.java
```diff
@@ -0,0 +1,11 @@
+package uk.ac.ox.ndph.mts.init_service.model;
```
alexbaxter, line 1: Is this class really necessary? It's just encapsulating a string really.

init-service/src/test/java/uk/ac/ox/ndph/mts/init_service/service/PractitionerServiceTest.java
```diff
@@ -7,18 +7,24 @@
...
 import uk.ac.ox.ndph.mts.init_service.exception.DependentServiceException;
+import uk.ac.ox.ndph.mts.init_service.model.IDResponse;
 import uk.ac.ox.ndph.mts.init_service.model.Practitioner;
+import uk.ac.ox.ndph.mts.init_service.model.RoleAssignment;
 
 import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
 
 import static org.assertj.core.api.Assertions.fail;
```
alexbaxter, line 22: Probably don't want to mix your unit test libraries here - use org.junit.jupiter.api.Assertions#fail() and #assertNotNull - won't change meaning of test

alexbaxter (commented):
Let me know if you want to make the suggested changes and I will approve this PR - they aren't significant.

init-service/src/test/java/uk/ac/ox/ndph/mts/init_service/service/PractitionerServiceTest.java
KateSant: Thanks, good spot, I've changed this.

init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/model/IDResponse.java
KateSant: It's for deserializing the bit of json returned by site-service and practitioner-service following a successful create.   (NB role service returns the whole role).    webclient has a convenient  .bodyToMono([class])  method which I'm using here sendBlockingPostRequest method here https://github.com/NDPH-ARTS/mts-services/blob/c62f75849f7f38b4a03685fbd4c19bda35b77e77/init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/service/ServiceInvoker.java and passing it this model.  Could possibly do the same with an anon inner class or something, but not sure it would be clearer?

alexbaxter (approved):
OK to merge.

darrengsmithson (approved):

PR #128, opened 2021-02-14
Smaller docker-compose footprint

tamirkamara:
## Description
Make some changed to lower the requirements of running the system via docker-compose, plus some other small fixes

### Changes
- Use a single SQL server in docker-compose
- Move containers to a smaller java base image
- Maven config to retry on download failure
- Use docker buildkit in CI builds
- Run init-service in ci integration test

### Checklist:

### DIFF
--- a/discovery-server/Dockerfile
+++ b/discovery-server/Dockerfile
@@ -21,14 +21,14 @@ COPY ./role-service/pom.xml ./role-service/
 COPY ./site-service/pom.xml ./site-service/
 
 # cache most dependencies as a layer to be used in other builds while POMs don't change
-RUN mvn dependency:go-offline --batch-mode -q --projects ${SVC}
+RUN mvn dependency:go-offline --batch-mode -q -Dmaven.wagon.http.retryHandler.count=3 --projects ${SVC}
 
 # copy current service source
 COPY ./${SVC}/src ${SVC}/src
 
 RUN mvn package --projects ${SVC}
 
-FROM openjdk:11.0.9.1-jre-buster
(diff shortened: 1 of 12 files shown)


shajidissa (dismissed):

eladiw (approved):

discovery-server/Dockerfile
```diff
@@ -21,14 +21,14 @@ COPY ./role-service/pom.xml ./role-service/
...
-RUN mvn dependency:go-offline --batch-mode -q --projects ${SVC}
+RUN mvn dependency:go-offline --batch-mode -q -Dmaven.wagon.http.retryHandler.count=3 --projects ${SVC}
 
 # copy current service source
 COPY ./${SVC}/src ${SVC}/src
 
 RUN mvn package --projects ${SVC}
 
-FROM openjdk:11.0.9.1-jre-buster
+FROM adoptopenjdk/openjdk11:jre-11.0.10_9-debianslim
```
alexbaxter, line 31: Can we set the JDK (and other image) tags in some base config for Docker, so they don't have to be repeated in different files?

alexbaxter (approved):
I'd prefer if there was less duplication in the docker files (per service), but I don't know how to modularize this. I've checked out the branch and it builds and starts up successfully and faster than before (not sure how much quicker); init-service still runs as well.

discovery-server/Dockerfile
tamirkamara: I don't think there's a concept of parent docker files. We have talked about putting in just 1 file in the root folder to build it all, and I've tried it once. I will try to bring that change too in the next few days.

PR #133, opened 2021-02-15
[ARTS-212] Assign Role to a Person Automation API tests

sameerapurini:
## Description
The PR consists of api automation tests to check the positive response when a user assigns a role to a person in a particular site. The role has permissions added too using permissions endpoint. The data is pulled dynamically from the data js file added within the tests folder. 

I have also moved the global utilities file 'utils' to the root of api-tests hence had to update the path of it in role service and permission files.

And also as the site name and alias fields doesn't accept numbers but only alphabets I had to update the 'getRandomString' function in the 'utils' accordingly.

### Changes
- [ARTS-212] - api tests added to cover assign role endpoint
- [ARTS-595] - Test approach

### Checklist:

[ARTS-212]: https://ndph-arts.atlassian.net/browse/ARTS-212

[ARTS-595]: https://ndph-arts.atlassian.net/browse/ARTS-595


api-tests/data/assign-roles/assign-roles.js
```diff
@@ -0,0 +1,39 @@
...
+const createPerson = {
+    "prefix": "Ms",
+    "givenName": "Cookie",
+    "familyName": "Cookie"
+}
+
+const createSite = {
+    "name": utils.getRandomString(5),
+    "alias": utils.getRandomString(3),
+    "parentSiteId": "5d01f252-d41c-40f8-b3c1-4bd7282fb28e",
```
alexbaxter, line 24: Where are you getting this ID from? Should be dynamically generated at init time?

api-tests/specs/assign-roles/assign-roles.js
```diff
@@ -0,0 +1,31 @@
+const requests = require('../../data/assign-roles/assign-roles')
+const conf = require('../../config/conf')
+const sitesEndpointUri = ':8083/sites';
```
alexbaxter, line 3: Can you get these service URLs from the config server? These are the local configurations I think, won't work in Azure maybe?

alexbaxter (commented):
No changes but a couple of questions.

api-tests/data/assign-roles/assign-roles.js
sameerapurini: that's right @alexbaxter I was about to come to you with regards to this. I am not sure if the 'init-services' will be run in the pr pipeline for me to pick it up from there. Or else I can just add another line of code to handle this.

alexbaxter: If you can reasonable assume the root site has been created, then do a GET on /sites and parse the ID out of that.

api-tests/specs/assign-roles/assign-roles.js
sameerapurini: I might need help with this @alexbaxter

alexbaxter (approved):
Approved, questions answered.

api-tests/specs/assign-roles/assign-roles.js
```diff
@@ -0,0 +1,44 @@
...
+const permissionsEndpointUri = ':8082/roles';
```
tamirkamara, line 4: Can you rename this to be "roles" endpoint?
Permissions are something different and might be confusing here.

tamirkamara (changes requested):

alexbaxter (dismissed):
Approved

tamirkamara (approved):

alexbaxter (approved):
Approved

PR #134, opened 2021-02-16
[ARTS-360] Validate sites in authorisation

LizaShak:
## Description
Validate the user site is authorised on entity's site. 

### Changes
- [ARTS-360] - The following changes were introduced:
1. Create a site-service client to get all sites
2. Build site sub trees ( used for inheritance check)
3. Validate the entity's site is in the user subtree 

### Checklist:

[ARTS-360]: https://ndph-arts.atlassian.net/browse/ARTS-360


sample-service/src/main/java/uk/ac/ox/ndph/mts/sample_service/security/authorisation/AuthorisationService.java
```diff
@@ -20,24 +28,46 @@
...
+                                final RoleServiceClient roleServiceClient,
+                                final SiteServiceClient siteServiceClient) {
         this.securityContextUtil = securityContextUtil;
+        this.siteTreeUtil = siteTreeUtil;
         this.practitionerServiceClient = practitionerServiceClient;
         this.roleServiceClient = roleServiceClient;
+        this.siteServiceClient = siteServiceClient;
     }
 
+    public boolean authorise(String requiredPermission, List<Object> sites, String methodName) {
```
alexbaxter, line 50: Presumably this method is the entry point for services wanting to check that the user has access to the required permission at the given sites? In that case accepting an arbitrary method name is a bit dangerous; also the list should be either a list of something that extends an interface e.g. List<HasSite> for somethings that can return a site (or site ID if they just return the ID). Then there is no need for callers to pass the methodName. If you do need to pass an arbitrary 'site ID getter' in here, prefer to use a lambda over method introspection i.e. signature should be:

    public boolean authorise(String requiredPermission, List<? extends Object> sites, Function<Object, String> idGetter) {...

but prefer: 

    public boolean authorise(String requiredPermission, List<? extends HasSiteId> sites) {...}

and 

```
public interface HasSiteId {
    String getSiteId();
}
```

sample-service/src/main/java/uk/ac/ox/ndph/mts/sample_service/security/authorisation/AuthorisationService.java
```diff
@@ -109,14 +140,29 @@ private boolean hasRequiredPermissionInRole(RoleDTO role, String requiredPermiss
...
-    @SuppressWarnings("squid:S1172") //suppress unused parameter
-    private boolean isSiteAuthorised(String entitySite, String userRole) {
-        return true;
+    private boolean isAuthorisedToAllSites(List<SiteDTO> sites,
+                                          List<RoleAssignmentDTO> roleAssignments,
+                                          List<String> entitiesSiteIds) {
+
+        Map<String, ArrayList<SiteDTO>> tree = siteTreeUtil.getSiteSubTrees(sites);
+
+        var hasAnUnauthorisedSite = entitiesSiteIds.stream().anyMatch(siteId ->
```
alexbaxter, line 149: This will grow as O(n*m) with the number of sites in the tree and the number of sites being checked, because there are two linear searches - can get very inefficient with even a relatively small number of sites.

sample-service/src/main/java/uk/ac/ox/ndph/mts/sample_service/security/authorisation/AuthorisationService.java
```diff
@@ -109,14 +140,29 @@ private boolean hasRequiredPermissionInRole(RoleDTO role, String requiredPermiss
...
+                        .map(RoleAssignmentDTO::getSiteId).distinct()
+                        .noneMatch(roleSiteId ->
+                                tree.get(roleSiteId).stream()
+                                        .anyMatch(siteDTO -> siteDTO.getSiteId().equals(siteId)))
+        );
+
+        return !hasAnUnauthorisedSite;
+    }
+
+    private String getSiteIdFromObj(Object obj, String methodName) {
```
alexbaxter, line 160: Get rid of this and use lambdas instead, as above comment

sample-service/src/main/java/uk/ac/ox/ndph/mts/sample_service/security/authorisation/SiteTreeUtil.java
```diff
@@ -0,0 +1,61 @@
...
+import uk.ac.ox.ndph.mts.sample_service.client.dtos.SiteDTO;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+@Component
+public class SiteTreeUtil {
```
alexbaxter, line 14: Going on past experience this method - get accessible subtrees of the sites tree - could be used a lot - might be worth having an endpoint in site-service to return it (filtering using these methods). Not necessary right now.

alexbaxter (changes requested):
Please change the method of getting site IDs from arbitrary instances in AuthorisationService as mentioned; consider optimizing the search for unauthorised sites in isAuthorisedToAllSites() - current logic works so I will approve with this.

sample-service/src/main/java/uk/ac/ox/ndph/mts/sample_service/security/authorisation/AuthorisationService.java
LizaShak: I thought about it a lot actually when writing this piece of code, and I also prefer the interface implementation. But that will mean that most of the models in our services will have to implement that. Is it something we are ok with ?

alexbaxter: The models won't have to implement it directly, the services can always wrap the actual model in a facade that translates to site IDs (which is essentially what you are doing in the service). The typed alternative is to use the lambda like I said above, or a generic version:

`public <T> boolean authorise(String requiredPermission, List<T> sites, Function<T, String> idGetter) {...
`

so here T is the type you are trying to authorise for, and the idGetter is the functional (lambda) to get the ID given a T.

alexbaxter (dismissed):
OK to merge now - future story might be to change the annotation to use a method reference rather than the method name, not sure if that is possible though.

sample-service/src/main/java/uk/ac/ox/ndph/mts/sample_service/security/README.md
```diff
@@ -0,0 +1,111 @@
+# Security
```
tamirkamara, line 1: Should this be in the sample service?
If yes, maybe not under a subfolder but in the root of the service and call it security.md?

sample-service/src/main/java/uk/ac/ox/ndph/mts/sample_service/security/authorisation/AuthorisationService.java
```diff
@@ -20,24 +29,46 @@
...
+                                final RoleServiceClient roleServiceClient,
+                                final SiteServiceClient siteServiceClient) {
         this.securityContextUtil = securityContextUtil;
+        this.siteTreeUtil = siteTreeUtil;
         this.practitionerServiceClient = practitionerServiceClient;
         this.roleServiceClient = roleServiceClient;
+        this.siteServiceClient = siteServiceClient;
     }
 
+    public boolean authorise(String requiredPermission, List<?> sites, String methodName) {
```
tamirkamara, line 51: These are all public methods that would be used across the application. Please add javadoc for all explaining the purpose and parameters.

About methodName:
1. Perhaps rename it since it's related to get site id from list?
2. Add a comment to explain why is it a string and used via reflection

tamirkamara (changes requested):

sample-service/src/main/java/uk/ac/ox/ndph/mts/sample_service/security/README.md
LizaShak: No it shouldn't, now when merging with main I will move it to the security package

alexbaxter (dismissed):
Looks good to me.

alexbaxter (approved):
Approved

tamirkamara (approved):

PR #142, opened 2021-02-18
Single dockerfile

tamirkamara:
## Description
Our services are built in a similar way, and their docker files also look the same. This change removed the dockerfiles at the service level in favor of using a common one in the root of the project.

### Changes
- Single docker file
- Add an option to cache dependencies of all the projects as a single layer. This is useful for local builds but not turned on by default.

### Checklist:


shajidissa (approved):
Looks much cleaner now.

Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+# dockerignore should keep the build context focused on what's required.
+COPY . ./
+RUN mvn package --projects ${SVC} --batch-mode -q --also-make -Dmaven.wagon.http.retryHandler.count=3
+
+FROM adoptopenjdk/openjdk11:jre-11.0.10_9-debianslim
+ARG SVC
+ARG JAR_FILE=${SVC}/target/*.jar
+COPY --from=builder-cached ${JAR_FILE} app.jar
+ENTRYPOINT ["java","-jar","/app.jar"]
+EXPOSE 8080:80
```
eladiw, line 49: need to verify docker-compose still works as each docker file (or most) had explicit ports (other than 8080)

mdp0 (approved):

eladiw (approved):

PR #144, opened 2021-02-19
Feature/arts 353 auth token

darrengsmithson:
## Description
Changes to allow authentication checks against an app service managed identity. The identity is  set as an env var when services are created via TF.

### Changes
- [ARTS-353] - authenticate  services
- Added authentication classes to the security package, updated the Main controller in the sample-service to check for authentication.
- Created a test for the  securityConextUtil class using Powermock for static mocking, therefore added Powermock dependencies to the pom
- Moved the SecurityContextUtil class to the authentication package as I think its more suited to authentication rather than authorization.

### Checklist:

[ARTS-353]: https://ndph-arts.atlassian.net/browse/ARTS-353


sample-service/src/main/java/uk/ac/ox/ndph/mts/sample_service/controller/MainController.java
```diff
@@ -44,7 +50,7 @@ public String getsecret() {
      *
      * @return a message as a string.
      */
-    @PreAuthorize("@authorisationService.authorise('stubPermission')") //NOSONAR
+    @PreAuthorize("@authenticationService.authenticate(env.getProperty(Managed_Identity)) or @authorisationService.authorise('stubPermission')") //NOSONAR
```
LizaShak, line 53: I think authenticate is a bit confusing here. That is after we have passed the authentication and now authorising the request... 
Also `env.getProperty(Managed_Identity) `and the:
`private final static String Managed_Identity = "MANAGED_IDENTITY";
`
introduces authorisation logic and authorisation properties to the service. So I think it is better to move it to the AuthorisationService

security/src/main/java/uk/ac/ox/ndph/mts/security/authentication/AuthenticationService.java
```diff
@@ -0,0 +1,34 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+
+/**
+ * The service which performs the authorisation flow
+ */
+@Service
+public class AuthenticationService {
```
LizaShak, line 13: Same for the authentication part... this is when we authorise the request...
I think we can just add a method to AuthorisationService something like:
`public boolean authoriseWithManagedIdentity()  {
        String identity = env.getProperty(Managed_Identity)
        return identity.equals(securityContextUtil.getUserId());
    }`

security/src/main/java/uk/ac/ox/ndph/mts/security/authentication/SecurityContextUtil.java
```diff
@@ -1,4 +1,4 @@
-package uk.ac.ox.ndph.mts.security.authorisation;
```
LizaShak, line 1: The SecurityContext is used to store the details of the _currently authenticated user_. It is when we already passed the authentication. So it should be under the authorisation.

LizaShak (changes requested):
I didn't write it everywhere. But it seems that we don't change anything on the authentication part. Just adding a new way to authorise a request for a token coming from the managed identity.

sample-service/src/main/java/uk/ac/ox/ndph/mts/sample_service/controller/MainController.java
darrengsmithson: this if for a service to service call so we havent 'passed' authentication for a service call - I agree with you that the code introduces authentication logic into the service and Id prefer for it not to happen but to 'inject' the MSI env var into the authentication service we would need to have the authentication service as an app in azure so TF can configure it. We cant do that if teh authentication service is in a jar... I think...

security/src/main/java/uk/ac/ox/ndph/mts/security/authentication/AuthenticationService.java
darrengsmithson: as above - if we use TF to inject the MSI value we would nee dot change to a service, I think? Otherwise we hard code property values in to property files in the security package ..which means rebuild/deploy when the MSI changes...

security/src/main/java/uk/ac/ox/ndph/mts/security/authentication/SecurityContextUtil.java
darrengsmithson: at the moment it just deals with the authentication?  

getUserId() {
        Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
        UserPrincipal userPrincipal = ((UserPrincipal) authentication.getPrincipal());

??
So it should be in authentication?

Authorization and other classes may use it - but its concerns are authentication?

sample-service/src/main/java/uk/ac/ox/ndph/mts/sample_service/controller/MainController.java
darrengsmithson: misunderstood the approach - Ive updated the code to remove the MSI being 'passed' to the auth service via the calling service.

security/src/main/java/uk/ac/ox/ndph/mts/security/authentication/AuthenticationService.java
darrengsmithson: as above, misunderstood the approach - changed code.

security/src/test/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationServiceTests.java
```diff
@@ -33,19 +36,36 @@
...
+        //Assert
+        assertTrue(authorisationService.authorise("some-permission"));
+    }
+
+
     @Test
     void TestAuthorise_WithInvalidToken_ReturnsFalse(){
         //Arrange
-        when(securityContextUtil.getUserId()).thenThrow(new AuthorisationException("Any exception during getting user from token"));
+        when(securityContextUtil.getUserId()).thenThrow(new AuthenticationException("Any exception during getting user from token"));
```
tamirkamara, line 68: I know you didn't add this, but seems like a redundant test to me - if the token is invalid how would we get passed the authentication of spring security...
This also makes the AuthenticationException (or AuthZ) class kind of mute.
@LizaShak?

tamirkamara (changes requested):
I think this change is inline with I'd originally thought we'll do. The only thing is to hide (if possible) the usage of managed identity in the code - i.e. not use it so much...

security/src/main/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationService.java
```diff
@@ -24,6 +27,9 @@
     private final PractitionerServiceClient practitionerServiceClient;
     private final RoleServiceClient roleServiceClient;
 
+    @Value("Managed_Identity")
+    private String managedIdentity;
```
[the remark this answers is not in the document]
darrengsmithson: yep, I dont disagree, Ill change and test.

security/src/main/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationService.java
```diff
@@ -43,6 +49,12 @@ public boolean authorise(String requiredPermission)  {
             //Get the user's object id
             String userId = securityContextUtil.getUserId();
 
+            //Managed Service Identities represent a call from a service and is
+            //therefore authorized.
+            if (isUserAManagedServiceIdentity(userId)) {
```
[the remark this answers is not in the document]
darrengsmithson: not sure, I thought about this a little while and changed it a few times...Im happy to change but not sure if isPreConfiguredServiceIdentity is better.. :)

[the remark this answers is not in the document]
darrengsmithson: the reason I kept it as isUserAManagedServiceIdentity is that I liked it when used in the 'if' statement with userId as a param....
if (isUserAManagedServiceIdentity(userId)) {
                return true;
            }
not sure if
if (isPreConfiguredServiceIdentity(userId)) {
                return true;
            }
is much better?? Dunno, Ill change if youre convinced..

security/src/test/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationServiceTests.java
darrengsmithson: I think Liza may have added it to ensure that Spring return a 403 due to forbidden access rules we throw an authenticationException rather than a generic Spring exception. Ill leave this in for the time being as the change/discussion is outside the scope of ARTS-353.

security/src/main/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationService.java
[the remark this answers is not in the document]
darrengsmithson: Ive changed the env var to init-service.identity but left the field name as managedIdentity. This is to take into consideration that its possible we may have have additional future managedIdentites from other services e.g another-service.identity, if this happens we convert the field name to private List<String> managedIdentities.

LizaShak (approved):

sample-service/src/main/java/uk/ac/ox/ndph/mts/sample_service/controller/MainController.java
```diff
@@ -13,6 +13,11 @@
 @RestController
 public class MainController {
 
+    private ConfigService myConfig;
```
eladiw, line 16: I think the var name should be changed myConfig -> something else . maybe configService?

darrengsmithson: fair point although its been like this for a while and its only in the sample service so not sure we need to change?

PR #148, opened 2021-02-23
[ARTS-609] Create person with auth

LizaShak:
## Description
Integrate the security package into practitioner service and add authorisation to create person endpoint.

### Changes
- [ARTS-609] - 
- Integrate security package into practitioner service
- Disable validation that on each endpoint we have authorisation annotation (because currently we added it just to one endpoint - create person, once we have fully integrated auth we can return this validation)
- Add authorisation to create person endpoint.

### Checklist:

- [ ] Branch name follows convention (feature/arts-#-short-name OR fix/arts-#-short-name)
- [ ] I have included a short-meaningful title, description and changes for this PR

PR #150, opened 2021-02-23
[ARTS-477] Bootstrap user login - profile endpoint

KateSant:
## Description
New /profile endpoint looks up practitioner data by accountID.
AccountId is from auth token using security module.
New FhirRepo and Service methods to do lookup.
Security config to allow OPTIONS pre-flight requests and CORS from the browser.

Azure deployment: https://sakate477uidev.z33.web.core.windows.net/

### Changes
- [ARTS-477] , subtask [ARTS-547]

### Checklist:

[ARTS-477]: https://ndph-arts.atlassian.net/browse/ARTS-477

[ARTS-547]: https://ndph-arts.atlassian.net/browse/ARTS-547


security/src/main/java/uk/ac/ox/ndph/mts/security/authentication/AADWebSecurityConfig.java
```diff
@@ -43,6 +43,8 @@ protected void configure(HttpSecurity http) throws Exception {
         // disable csrf because we are using another token mechanism
         http.csrf().disable(); //NOSONAR
 
+        http.cors(); // See https://www.baeldung.com/spring-cors
```
KateSant, line 46: @LizaShak and @mdp0 I want to draw your attention this since it's a change to the core security package.   It's needed for story 477 but it's really a wider issue for any UI stories.

security/src/main/java/uk/ac/ox/ndph/mts/client/practitioner_service/PractitionerServiceClient.java
```diff
@@ -9,6 +9,6 @@
  */
 public interface PractitionerServiceClient {
 
-    List<RoleAssignmentDTO> getUserRoleAssignments(String userId);
+    List<RoleAssignmentDTO> getUserRoleAssignments(String userId, String token);
```
alexbaxter, line 13: Can the token be injected by Spring rather than as a parameter? We will have it everywhere if not.

alexbaxter (commented):
Not for merge yet - need to look at if we can inject the token rather than passing it as a parameter - maybe one for Tamir's common service client library

security/src/main/java/uk/ac/ox/ndph/mts/client/practitioner_service/PractitionerServiceClient.java
KateSant: This is one of the commits from story 609 / PR 148, not this story.   Do you mind if we move this comment to https://github.com/NDPH-ARTS/mts-services/pull/148 so that @LizaShak can review it.  It will be confusing if we do it here.

docker-compose.yml
```diff
@@ -21,9 +21,11 @@ services:
       FHIR_URI: "http://fhir-api:8099"
       ROLE_SERVICE_URI: "http://role-service:8082"
       SITE_SERVICE_URI: "http://site-service:8083"
+      PRACTITIONER_SERVICE_URI: "http://practitioner-service:8081"
```
eladiw, line 24: does the practitioner needs its own uri?

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/controller/PractitionerController.java
```diff
@@ -16,36 +19,41 @@
...
 
+import static org.springframework.http.HttpStatus.CREATED;
+import static org.springframework.http.HttpStatus.OK;
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+
 /**
  * Controller for practitioner endpoint of practitioner-service
  */
 @RestController
+@CrossOrigin(origins = "http://localhost:4200")  // See https://www.baeldung.com/spring-cors
```
eladiw, line 36: which endpoint is 4200? is it just for local work? or is it also when deployed on azure?

alexbaxter (approved):
No change since list review, still OK to merge

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/controller/PractitionerController.java
KateSant: Yes this is for local work - the Angular front end runs on port 4200 in the dev environment.  This means that when the front end makes an http request to the back end, the browser thinks it is a Cross Origin Request and so makes a pre-flight options request.   This annotation tells Spring to give the correct headers to make the browser happy.

alexbaxter (dismissed):
Looks good

docker-compose.yml
KateSant: The practitioner service has its own URI.  But following recent merges there are no changes to docker-compose in this PR.

init-service/src/main/resources/application-local.yml
```diff
@@ -46,7 +46,7 @@ trial:
       prefix: Mr
       roles:
         - superuser
-      userAccount: some-user-account-id
+      userAccount: 5d995fd0-a1ff-4b2c-ba93-238ba9e349b1
```
alexbaxter, line 49: Is that a secret? Should it go in the config?

alexbaxter (commented):
See comment on application-local.yml

init-service/src/main/resources/application-local.yml
KateSant: I don't think it's a secret. It's Mr Bootstrap's AD ID - he only exists for the purpose of proving story 477.

alexbaxter: So should come from the config-server right?

KateSant: We found another way to approach this so I'm taking it out.

alexbaxter (dismissed):
Approving, Kate to add a new issue for getting the bootstrap user-id from the configuration server (trial config).

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/controller/PractitionerController.java
LizaShak: Hi, following our conversation, we need to test it on Azure see it works there ...

LizaShak (changes requested):

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/controller/PractitionerController.java
KateSant: It didn't work on Azure.  I discussed this with Mark and we decided to permit CORS from all origins.  Security issue does not arise given the auth we have in place.  (NB Liza, that other problem with logging into the Angular UI was not to do with injection of the clientid after all - Elad helped me eliminate that. The problem was a bug in the Angular config, which Mark was aware of and has already fixed it. his PR for that fix is blocked by a webdriver deployment issue but he's working on that now. We tested it with a workaround.)

KateSant: I have moved the CORS config into the gateway.  Here it is working on Azure https://sakate477uidev.z33.web.core.windows.net/

alexbaxter (approved):
Approved

gateway-server/src/main/resources/bootstrap.yml
```diff
@@ -32,6 +32,15 @@ spring:
             - Path=/api/roles/**
           filters:
             - StripPrefix=1
+      globalcors:
+        corsConfigurations:
+          '[/**]':
+            allowedOrigins: "*" # TODO : limit origins - or reconsider CORS approach?
```
LizaShak, line 38: Maybe it is worth adding a technical story for that...

LizaShak (approved):

gateway-server/src/main/resources/bootstrap.yml
KateSant: Good idea - I will raise a Jira ticket

PR #165, opened 2021-03-04
[ARTS-364] Add Application Insights for logs and metrics

tamirkamara:
## Description
This adds application insights to all services to collect logs and metrics.
This can be merged without waiting for the deployment related PR.

### Changes
- [ARTS-364] Collect logs/metrics
- Add spring cloud config "fail-fast" because if the config isn't available at startup, no point to bring up the/any service.

### Checklist:

[ARTS-364]: https://ndph-arts.atlassian.net/browse/ARTS-364

### DIFF
--- a/applicationinsights.json
+++ b/applicationinsights.json
@@ -1,0 +1,13 @@
+{
+  "sampling": {
+    "percentage": 100
+  },
+  "instrumentation": {
+    "logging": {
+      "level": "DEBUG"
+    },
+    "micrometer": {
+      "enabled": true
+    }
+  }
+}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,16 @@
 # This dockerfile includes "conditional layers" and best used with DOCKER_BUILDKIT=1.
-
 ARG DEPS_CACHE=project
+ARG BUILDER_IMAGE=maven:3-openjdk-11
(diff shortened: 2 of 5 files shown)


applicationinsights.json
```diff
@@ -0,0 +1,13 @@
...
+  },
+  "instrumentation": {
+    "logging": {
+      "level": "DEBUG"
+    },
+    "micrometer": {
+      "enabled": true
+    }
+  }
+}
```
alexbaxter, line 13: What kind of impact on performance will this have (if any)? i.e. does this need to be turned off for 'production' use, or is it safe to use the same configuration for all instances?

alexbaxter (approved):
Output posted by Tamir looks good so approving.

applicationinsights.json
mdp0: They look like parameters we'd want to be able to adjust for sure - are they configurable through environment variables like some of the other app insights params?

tamirkamara: This can be also set by an Env var like documented here: https://docs.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-config#auto-collected-logging
I can remove it here and push it through the deployment or we can leave that for later?

mdp0: I think this is good for a starter, the controls are available to dial it back if we need to later on.

mdp0 (approved):

eladiw (approved):

PR #169, opened 2021-03-05
[ARTS-621] assign role with authorisation

darrengsmithson:
## Description
Added authorisation to assign roles - an update to the init-service application.yml property files in mts-trial-deployment-config repo is needed to test this. API tests will fail unless 
a)docker-compose points at branch - https://github.com/NDPH-ARTS/mts-trial-deployment-config/tree/ARTS-621-assign-role-auth - or 
b)we commit the config to main with..
https://github.com/NDPH-ARTS/mts-trial-deployment-config/pull/70

### Changes
- [ARTS-621] - Assign roles with authorisation
- Added auth to controller endpoint
- Updated tests

### Checklist:

[ARTS-621]: https://ndph-arts.atlassian.net/browse/ARTS-621


LizaShak (changes requested):
This endpoint is also used by the init service, therefore we should have a pr in mts-trial-deployment-config  repo where we add it, and make sure the init service runs and this feature.works on Azure

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/controller/PractitionerController.java
```diff
@@ -67,6 +67,7 @@ public PractitionerController(EntityService entityService) {
         return ResponseEntity.ok(entityService.findPractitionerById(id));
     }
 
+    @PreAuthorize("@authorisationService.authorise('assign-role')") //NOSONAR
```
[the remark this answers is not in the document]
darrengsmithson: Updated - also added extra auth service mocking incase other services copy and paste and it uses other methods..

LizaShak (approved):

LizaShak (approved):

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/controller/PractitionerController.java
```diff
@@ -71,6 +71,7 @@ public PractitionerController(EntityService entityService,  SecurityContextUtil
         return ResponseEntity.ok(entityService.findPractitionerById(id));
     }
 
+    @PreAuthorize("@authorisationService.authorise('assign-role', #roleAssignment.siteId)") //NOSONAR
```
KateSant, line 74: Are you sure you need a //NOSONAR annot?  I don't think sonar complains about preauth annotations?

security/src/main/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationService.java
```diff
@@ -121,13 +121,15 @@ public boolean authorise(String requiredPermission, List<String> entitiesSiteIds
...
             }
 
-            if (!hasValidPermissions(requiredPermission, roleAssignments)) {
+            var rolesAssignmentsWithPermission = getRolesAssignmentsWithPermission(requiredPermission, roleAssignments);
+
+            if (rolesAssignmentsWithPermission.isEmpty()) {
                 return false;
             }
 
             List<SiteDTO> sites = siteServiceClient.getAllSites();
```
KateSant, line 130: Just getting my head around the logic and the rest calls here - do I have this right?
(1) get all role assignments for practitioner P   (from practitioner-service)
(2) get permissions for those roles (from role-service) 
(3) filter role assignments to remove those which connect to roles which don't connect to permission X 
(4) get all sites (from site-service) and buildsite tree hierarchy
(5) filter role assignments again, this time by site S and child-sites CS.

I wondered whether we could do the site filtering at an earlier stage, e.g. by asking practitioner-service only for role assignments at site S, for practitioner P.  But I don't think that can work unless practitioner-service knows the site-tree hierarchy, otherwise it will miss child sites is that right?   

I wondered about performance here with so many rest calls.  I think the logic is necessary to achieve the story.  So my only thought is - maybe cache some of the data?

LizaShak: You are right , you can't because of the child hierarchy. Regarding caching, we thought about it during the design too. You can cache the whole site tree. I'll check if there is already a story for that.

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/controller/PractitionerController.java
LizaShak: yes you have too, because sonar recognises it as a security issue, it ask to use role based access or other pre known validation. Since we are using custom authorisation it doesn't recognise it.

KateSant (dismissed):

shajidissa (approved):

sameendph (approved):
approving

PR #183, opened 2021-03-09
[ARTS-622] Authz on create site endpoint 

KateSant:
## Description
* Integrate security component into site-service
* New create-site permission in role-service liquibase config
* Add authorization annotation to the create site endpoint - referencing parent siteid
* init-service sends token to site-service when posting to create site endpoint
* authz component sends token to site-service when getting sites
* Null check in authz component 

Verified with postman see: https://ndph-arts.atlassian.net/browse/ARTS-622?focusedCommentId=10610

### Changes
- [ARTS-622] 

### Checklist:

[ARTS-622]: https://ndph-arts.atlassian.net/browse/ARTS-622


practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/client/AbstractEntityServiceClient.java
```diff
@@ -11,13 +12,17 @@
...
+
+
 
     @SuppressWarnings("ConstantConditions")
     @Override
     public boolean entityIdExists(String id) throws RestException {
         Objects.requireNonNull(id, "id must be non-null");
         return webClient.get()
                 .uri(serviceExistsRoute, id)
+                .header("Authorization", "Bearer " + securityContextUtil.getToken())
```
LizaShak, line 25: I think it is better to use the constant:
`.header(HttpHeaders.AUTHORIZATION, "Bearer " + token)`

Same in other calls...

LizaShak (approved):
Approved with minor comment

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/client/AbstractEntityServiceClient.java
KateSant: Thanks Liza - good point.   I have changed this here in the practitioner service and also in security/src/main/java/uk/ac/ox/ndph/mts/client/site_service/SiteServiceClientImpl.java

alexbaxter: If you use the headers() method you can use:
`.uri(...route..).headers( (headers) -> headers.setBearerAuth(token) )`
- or pass a `Consumer<Headers>` customizer - see role-service-client

security/src/main/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationService.java
```diff
@@ -99,20 +100,30 @@ public boolean authorise(String requiredPermission) {
...
+
         try {
             //Get the user's object id
             String userId = securityContextUtil.getUserId();
             String token = securityContextUtil.getToken();
 
-            LOGGER.debug("userId is - " + userId);
-            LOGGER.debug("managed identity is - " + managedIdentity);
+            LOGGER.debug("userId Is - {}", userId);
+            LOGGER.debug("managed identity is - {}", managedIdentity);
```
alexbaxter, line 110: Debug code - remove unless wanted for debugging in QA

alexbaxter (dismissed):
Approved - leave in debug logs for debugging latest version of authentication code

security/src/main/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationService.java
KateSant: We discussed in a different thread and decided to leave it in as debugging for now while the managed service identity is still being worked on (not in this story).

darrengsmithson (commented):
All changes are good - for the existing code...just wonder if using a default method in the EntityServiceClient for entityIdExists that houses the code would mean that we dont need the abstract class? Also maybe enums rather than hard-coded Strings..nut these are refactorings rather than review comments

KateSant:
> All changes are good - for the existing code...just wonder if using a default method in the EntityServiceClient for entityIdExists that houses the code would mean that we dont need the abstract class? Also maybe enums rather than hard-coded Strings..nut these are refactorings rather than review comments

Thanks Darren I agree with the refactor you are proposing on practitioner-service - we could collapse AbstractEntityServiceClient and EntityServiceClient interface.   However, since all of the service-to-service client code is going to be replaced with a common library ARTS-636, I think it might be wasted work?  Also this story is primarily about site-service so I'm a bit hesitant to trample more deeply into practitioner-service than I already have?

darrengsmithson (approved):

LizaShak (dismissed):

LizaShak (dismissed):

darrengsmithson (approved):

shajidissa (approved):

PR #197, opened 2021-03-17
[ARTS-793] - Increasing default search result set from FHIR

darrengsmithson:
Increasing result set size form FHIR to 50

## Description
We are currently only returning a max of 10 results from FHIR. Increasing result set size form FHIR to 50.

This change is linked to the config repo changes and therefore API tests will fail until the linked PR is merged into main - https://github.com/NDPH-ARTS/mts-trial-deployment-config/pull/87/files

### Changes
- [ARTS-793] - Increasing result set size form FHIR to 50

### Checklist:

[ARTS-793]: https://ndph-arts.atlassian.net/browse/ARTS-793

### DIFF
--- a/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/repository/FhirContextWrapper.java
+++ b/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/repository/FhirContextWrapper.java
@@ -18,7 +18,7 @@ import java.util.List;
 @Component
 public class FhirContextWrapper {
     private final FhirContext fhirContext;
-    private final int RESULT_COUNT = 50;
+    private final int RESULTS = 50;
 
     /**
      * constructor
@@ -87,7 +87,7 @@ public class FhirContextWrapper {
         return fhirContext.newRestfulGenericClient(uri)
                 .search()
                 .forResource(resourceClass)
-                .count(RESULT_COUNT)
+                .count(RESULTS)
                 .returnBundle(Bundle.class);
     }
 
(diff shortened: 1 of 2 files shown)


site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/repository/FhirContextWrapper.java
```diff
@@ -18,6 +18,7 @@
 @Component
 public class FhirContextWrapper {
     private final FhirContext fhirContext;
+    private final int RESULT_COUNT = 50;
```
KateSant, line 21: Is 50 the right number - might we have more sites/practitioners/whatevers than that?  (Is there any performance penalty if we give a high number here?)

shajidissa (dismissed):

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/repository/FhirContextWrapper.java
tamirkamara: Might be worth to externalize as a config setting to control more easily.

darrengsmithson: not sure - just put an arbitrary larger number than 10 - can up to 100 if you want - the amount of data being passed in 1 request rather than multi requests via paging or similar hasnt been clarified etc.. Im sure as the app grows this is something that will be discussed and therefore functionality like this will also change...thinking about it, it maybe best to keep it on the lower side so not to mask any perf issues..if we start getting more than 50 results that might push us into a perf type discussion??

darrengsmithson: fair point bout the config approach - Ill update

darrengsmithson: updated config files and set the default value to 50 - we can easily adjust as per config changes if required - good catch Kate/Tamir..

KateSant (approved):

shajidissa (approved):

PR #199, opened 2021-03-18
[ARTS-207] - API tests covering address fields at sites endpoint

sameerapurini:
The PR contains a set of scenarios to check the presence of an address at CCO and LCC level. Negative scenarios added to see if the address field present for non-configured sites like regions and countries fails.
Also updated the author name in the api test files that would help in tracking down the QA.

Changes

ARTS-207 -Address field added to the site with no validation

### Checklist:

### DIFF
--- a/api-tests/common/utils.js
+++ b/api-tests/common/utils.js
@@ -1,3 +1,10 @@
+/**
+ * Utils class class covering functions to use across the api tests
+ * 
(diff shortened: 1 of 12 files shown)


sameendph (approved):
approving

shajidissa (approved):

PR #200, opened 2021-03-18
Adding debug logs + Adding liquibase config

eladiw:
Adding some debug logs across the services + moved liquibase config to service itself (role).

### Checklist:

### DIFF
--- a/init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/loader/Loader.java
+++ b/init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/loader/Loader.java
@@ -15,7 +15,7 @@ import java.util.List;
 
 @Component
 public class Loader implements CommandLineRunner {
-    private static final Logger LOGGER = LoggerFactory.getLogger(CommandLineRunner.class);
+    private final Logger logger = LoggerFactory.getLogger(CommandLineRunner.class);
 
     private final PractitionerServiceInvoker practitionerServiceInvoker;
     private final RoleServiceInvoker roleServiceInvoker;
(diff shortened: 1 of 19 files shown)


init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/loader/Loader.java
```diff
@@ -13,6 +15,7 @@
 
 @Component
 public class Loader implements CommandLineRunner {
+    private final Logger logger = LoggerFactory.getLogger(CommandLineRunner.class);
```
darrengsmithson, line 18: use the more standard approach - private static final Logger LOGGER ?? Theres a little bit of a mix in the codebase tbf..

darrengsmithson (commented):
PR title?

darrengsmithson (commented):
I dont think the liquibase properties should be in the bootstrap file - the bootstrap should be used for loading configuration properties from external sources. I also think we're masking the real issue here - shouldnt we be always making the config server be available B4 other services start? Even if we workaround this issue in this service for this problem what if we need to config server to start 1st for another service - do we then 'move' those config properties to the bootstrap and thus the bootstrap becomes a fail-over option for non related properties??

init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/loader/Loader.java
eladiw: I moved everything to non  static, all except places that must have static loggers (i.e. static methods)

LizaShak (approved):

PR #205, opened 2021-03-21
[ARTS-729] Correct auth - read scope from config

LizaShak:
## Description
Read scope from configuration

### Changes
- [ARTS-729] - correct authentication flow

### Checklist:

[ARTS-729]: https://ndph-arts.atlassian.net/browse/ARTS-729

### DIFF
--- a/init-service/src/main/resources/application.yml
+++ b/init-service/src/main/resources/application.yml
@@ -1,9 +1,6 @@
 spring:
   main:
     web-application-type: NONE # init-service is just a console app, should quit when done.
-azure:
-  activedirectory:
-    scope: api://mts-dev/default
 role-service:
   uri: http://role-service
   routes:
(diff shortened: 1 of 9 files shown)


init-service/src/main/resources/application.yml
```diff
@@ -1,6 +1,9 @@
 spring:
   main:
     web-application-type: NONE # init-service is just a console app, should quit when done.
+azure:
+  activedirectory:
+    scope: api://mts-dev/default
```
tamirkamara, line 6: azure.activedirectory is something the azure libs are using, but this scope is an mts specific one.
Also, maybe worth to bring it from the config server and no to hard-code it here.

tamirkamara (changes requested):

init-service/src/main/resources/application.yml
darrengsmithson: agree, shouldnt this should be in config server?

darrengsmithson (changes requested):

PR #211, opened 2021-03-25
Init service starts only when ALL expected services are registered

eladiw:
## Description
Init service starts only when ALL expected services are registered.
In this way we improve the init service to succeed and not reach any undesired state due to partial initialization.

It can still happen (e.g. service is unregistered after the check was done), but it improves the chance of success.

### Changes
Discovery client was added to the init-service. We query it for the registered services and we block the init until we verify all are registered.

image

### Checklist:

### DIFF
--- a/init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/loader/Loader.java
+++ b/init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/loader/Loader.java
@@ -51,8 +51,8 @@ public class Loader implements CommandLineRunner {
         initProgressReporter.submitProgress(String.format(LoaderProgress.ENTRY_POINT.message(), delayStartInSeconds));
         Thread.sleep(delayStartInSeconds * 1000);
 
-        var allReady = false;
-        while (!allReady) {
+        boolean allReady;
+        do {
             initProgressReporter.submitProgress(LoaderProgress.WAITING_FOR_ALL.message());
 
             var applications = this.discoveryClient.getServices();
@@ -65,12 +65,10 @@ public class Loader implements CommandLineRunner {
 
             allReady = applications.containsAll(
                 Arrays.asList("config-server", "site-service", "role-service", "practitioner-service"));
-            if (allReady) {
-                break;
+            if (!allReady) {
+                Thread.sleep(5000);
             }
-
-            Thread.sleep(5000);
-        }
+        } while (!allReady);
 
         initProgressReporter.submitProgress(LoaderProgress.ALL_REGISTERED.message());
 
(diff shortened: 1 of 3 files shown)


init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/loader/Loader.java
```diff
@@ -43,9 +49,31 @@ public Loader(Trial trialConfig,
     @Override
     public void run(String... args) throws InterruptedException, IOException {
         // Give the other services some time to come online.
+        // We wait and query the discovery to see that our services are up and ready
         initProgressReporter.submitProgress(String.format(LoaderProgress.ENTRY_POINT.message(), delayStartInSeconds));
         Thread.sleep(delayStartInSeconds * 1000);
```
tamirkamara, line 54: Is this still required?

init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/loader/Loader.java
```diff
@@ -43,9 +49,31 @@ public Loader(Trial trialConfig,
...
 
+        var allReady = false;
+        while (!allReady) {
+            initProgressReporter.submitProgress("Waiting for all services to be registered");
```
tamirkamara, line 58: Can you move this string to the enum?

init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/loader/Loader.java
```diff
@@ -43,9 +49,31 @@ public Loader(Trial trialConfig,
...
+
+            var applications = this.discoveryClient.getServices();
+            initProgressReporter.submitProgress(applications.size() + " services are registered");
```
tamirkamara, line 61: Can you move this string to the enum?

init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/loader/Loader.java
```diff
@@ -43,9 +49,31 @@ public Loader(Trial trialConfig,
...
+            for (var application : applications) {
+                initProgressReporter.submitProgress(application + " is registered");
```
tamirkamara, line 63: Can you move this string to the enum?

init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/loader/Loader.java
```diff
@@ -43,9 +49,31 @@ public Loader(Trial trialConfig,
...
+            allReady = applications.containsAll(
+                Arrays.asList("config-server", "site-service", "role-service", "practitioner-service"));
+            if (allReady) {
+                break;
+            }
+
+            Thread.sleep(10000);
+        }
+
+        initProgressReporter.submitProgress("All services were registered. Continue.");
```
tamirkamara, line 75: Can you move this string to the enum?

tamirkamara (changes requested):

init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/loader/Loader.java
eladiw: As discussed. it is partially required. as in real life those services won't be available from the start (discovery also), so we give it a bit of grace (will reduce the 120 sec constant to a lower value)

tamirkamara (approved):

guybartal (approved):
Looks good!

PR #254, opened 2021-04-13
Feature/arts 774 fix permissions

darrengsmithson:
## Description
Missed permission check on 1st PR. Needed to add an auth endpoint in site controller to avoid an endless loop from site controller to auth service and back. This will probably need to be done in other services too but a seperate ticket can be raised for that.
Also added the status flag being returned.

### Changes
- [ARTS-774] - adding permission check
- [ARTS-76] - retrieving status

### Checklist:

[ARTS-774]: https://ndph-arts.atlassian.net/browse/ARTS-774

[ARTS-76]: https://ndph-arts.atlassian.net/browse/ARTS-76

### DIFF
--- a/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/controller/SiteController.java
+++ b/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/controller/SiteController.java
@@ -37,6 +37,7 @@ public class SiteController {
         return ResponseEntity.status(HttpStatus.CREATED).body(new Response(siteId));
     }
 
+    @PreAuthorize("@authorisationService.authorise('view-site')")
     @PostAuthorize("@authorisationService.filterUserSites(returnObject.getBody(), #role)")
     @GetMapping
     public ResponseEntity<List<Site>> sites(@RequestParam(value = "role", required = false) String role) {
--- a/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/controller/SiteController.java
+++ b/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/controller/SiteController.java
@@ -37,7 +37,6 @@ public class SiteController {
         return ResponseEntity.status(HttpStatus.CREATED).body(new Response(siteId));
     }
 
-    @PreAuthorize("@authorisationService.authorise('view-site')")
     @PostAuthorize("@authorisationService.filterUserSites(returnObject.getBody(), #role)")
     @GetMapping
     public ResponseEntity<List<Site>> sites(@RequestParam(value = "role", required = false) String role) {
--- a/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/controller/SiteController.java
+++ b/site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/controller/SiteController.java
@@ -44,8 +44,9 @@ public class SiteController {
         return ResponseEntity.status(HttpStatus.OK).body(siteService.findSites());
     }
 
-    @GetMapping("/authorize")
-    public ResponseEntity<List<Site>> authorizeSites(@RequestParam(value = "role", required = false) String role) {
+    @GetMapping("/unauthorized")
+    @PostAuthorize("@authorisationService.filterUserSites(returnObject.getBody(), null)")
+    public ResponseEntity<List<Site>> unauthorizedSites(@RequestParam(value = "role", required = false) String role) {
         return ResponseEntity.status(HttpStatus.OK).body(siteService.findSites());
     }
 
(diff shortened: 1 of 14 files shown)


site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/controller/SiteController.java
```diff
@@ -32,14 +32,20 @@ public SiteController(SiteService siteService) {
...
 
+    @PreAuthorize("@authorisationService.authorise('view-site')")
     @PostAuthorize("@authorisationService.filterUserSites(returnObject.getBody(), #role)")
     @GetMapping
-    public ResponseEntity<List<Site>> sites(@RequestParam(value = "role", required = false) String role) {
+    public ResponseEntity<List<Site>> getSites(@RequestParam(value = "role", required = false) String role) {
+        return ResponseEntity.status(HttpStatus.OK).body(siteService.findSites());
+    }
+
+    @GetMapping("/authorize")
```
KateSant, line 47: So this new GET /authorize endpoint does the same thing as the GET /sites endpoint (i.e. return all sites) but bypassing authorization?    Is this the right RESTful name for it - it's more un-authorized than authorize?   Also I wondered whether there is any alternative solution for the circularity problem in the authz component, other than a duplicate endpoint?  @LizaShak talked about this problem before but I can't remember what the conclusion was.

LizaShak, line 47: It still need the PostAuthorize header that was here before:
maybe something like:
@PostAuthorize("@authorisationService.filterUserSites(returnObject.getBody(), null)")
Because it should only return sites user has a role assignment on

LizaShak (changes requested):

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/controller/SiteController.java
LizaShak: In the endpoints called by the authorisation we can't call authorise. That will cause a loop. If you need an endpoint that requires a permission check... You add it on the endpoints used by authorisation.. So I guess it has to be a separate enpoint right?

KateSant: So this will be the case with all 3 endpoints called by the authz logic i.e. /sites for the site tree,  /practitioners/roles for role assignments and /roles?ids=1,2,3 for role-permission mappings?   All will need a duplicate endpoint bypassing the normal authz that would be applied to other callers?    How do we secure them?

darrengsmithson: ok..so when auth service calls the endpoint it could be from other endpoints that dont have the postAuthorize - if so, wouldnt it be better to add the filter method in authSevice rather than add an annotation which calls the auth service to the endpoint?

darrengsmithson: ok unauthorized as teh sites at that point arent authorized - I was thinking that I just wanted to make it clear that only the auth service should access the endpoint but happy to change - I think we'll end up refactoring anyhow

KateSant: Unless we can secure the duplicate endpoints, this solution is storing up a new problem?     I see that /practitioner/roles returns only the user's own roles which limits access.   But that won't work with the sites endpoint which needs to return the whole tree for the authz logic?   On the naming, /unauthorized seems confusing too now I look at it - I can't work out what the endpoint should be called and that's maybe a sign that something is up with this solution?    Would it be heresy for the authz component to get its data in another way?

LizaShak: So that is not what BAs wanted. Basically someone can just now call this endpoint and get all the sites, something I don't think we want to happen. Also I know remembered this endpoint is also used by the ui as part of story [ARTS-48](https://ndph-arts.atlassian.net/browse/ARTS-48) (by the way, this also means we need to change the path in the ui where it does the get my sites)

KateSant (approved):

LizaShak (commented):
Following our conversation unblocking my review.

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/controller/SiteController.java
darrengsmithson: changed endpoint to sites/assigned and added the postAuth to filter which now at least fits with what the endpoint does i.e return sites after post filtering on roleassignments.

KateSant (approved):

mdp0 (approved):

PR #257, opened 2021-04-15
fix/[ARTS-910] Role service local config

KateSant:
Local config to enable a developer to run role-service locally without a config server or dependency on the mts-trial-deployment-config repo

### Checklist:

### DIFF
--- a/role-service/src/main/resources/bootstrap-local.yml
+++ b/role-service/src/main/resources/bootstrap-local.yml
@@ -6,3 +6,11 @@ spring:
         enabled: false
     discovery:
       enabled: false
+  liquibase:
+    change-log: classpath:db/changelog/changelog-role-service-master.xml
+    drop-first: false
+    enabled: true
+  jpa:
+    hibernate:
+      ddl-auto: none
+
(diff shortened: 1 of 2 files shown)


darrengsmithson (changes requested):
We seem to have lost application-local.yml - I think Tamir may have removed it - I think I added a comment to his review...anyhow, if application-local.yml is in the resources folder bootstrap will pick up the liquibase properties.

mdp0 (dismissed):

KateSant:
> We seem to have lost application-local.yml - I think Tamir may have removed it - I think I added a comment to his review...anyhow, if application-local.yml is in the resources folder bootstrap will pick up the liquibase properties.

Thanks Darren, do you mean this config should go in an application-local.yml rather than bootstrap-local.yml?

role-service/src/main/resources/bootstrap-local.yml
```diff
@@ -6,3 +6,11 @@ spring:
         enabled: false
     discovery:
       enabled: false
+  liquibase:
```
tamirkamara, line 9: Can all these be moved to app local file?

KateSant: Done

darrengsmithson (approved):

mdp0 (approved):

PR #262, opened 2021-04-22
[ARTS 916] fix config for local discovery server 

KateSant:
## Description
Init-service waits for other services to appear in the discovery-server registry (See https://github.com/NDPH-ARTS/mts-services/blob/main/init-service/src/main/java/uk/ac/ox/ndph/mts/init_service/loader/Loader.java).   At the moment all services have discovery switched off in local profile.   So if we we want the option to run init-service locally, we need to run discovery locally, or do a workaround.

### Changes
- [ARTS-916] 

### Checklist:

[ARTS-196]: https://ndph-arts.atlassian.net/browse/ARTS-196

[ARTS-916]: https://ndph-arts.atlassian.net/browse/ARTS-916

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -51,9 +51,12 @@ services:
       SPRING_CLOUD_CONFIG_LABEL: "main"
       SPRING_CLOUD_CONFIG_URI: "http://config-server:8888"
       LOGGING_LEVEL_ROOT: ${LOGGING_LEVEL_ROOT:-INFO}
+      ROLE_SERVICE_URI: "http://role-service:8082"
     depends_on:
-      - config-server
-      - sql
+        config-server:
+          condition: service_healthy
+        sql:
+          condition: service_healthy
     healthcheck:
       test: curl --fail http://localhost:8082/actuator/health
       interval: 10s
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -51,6 +51,7 @@ services:
       SPRING_CLOUD_CONFIG_LABEL: "main"
       SPRING_CLOUD_CONFIG_URI: "http://config-server:8888"
       LOGGING_LEVEL_ROOT: ${LOGGING_LEVEL_ROOT:-INFO}
+      ROLE_SERVICE_URI: "http://role-service:8082"
```
darrengsmithson, line 54: I add the following to security application.yml - thats where the other services xx.service.uri details are..

role:
  service:
    name: role-service
    uri: http://role-service
    endpoint:
      sites: /roles

darrengsmithson (changes requested):

docker-compose.yml
KateSant: > I add the following to security application.yml - thats where the other services xx.service.uri details are..
> 
> role:
> service:
> name: role-service
> uri: http://role-service
> endpoint:
> sites: /roles

Good call thanks Darren - I've moved this to application.yaml with the others.

darrengsmithson (approved):

sameerapurini (approved):

PR #263, opened 2021-04-29
Feature/arts 915

darrengsmithson:
## Description
Please replace this text with a summary of the change and which issue is fixed.

### Changes
- [ARTS-915] - changes to the auth process, remove reflection, move some code to site package
- tidy up of code for readability
- introduce a SiteDTO for data transfer 
- update tests
### Checklist:

[ARTS-915]: https://ndph-arts.atlassian.net/browse/ARTS-915


api-tests/specs/createTrialSite/createNewTrialSite.js
```diff
@@ -155,7 +155,7 @@ describe('As a user with Create Trial Sites permission I want to have the system
             method: 'POST',
             body: JSON.stringify(requests.missingParent),
         })
-        expect(fetchResponse11.status).to.equal(HttpStatus.FORBIDDEN)
+        expect(fetchResponse11.status).to.equal(HttpStatus.UNPROCESSABLE_ENTITY)
```
darrengsmithson, line 158: while testing noticed that a 403 was being thrown when a parentSiteId was passed - this should be a 422 - I changed the code so that the error occurs when validating and not when doing auth.

practitioner-service/src/main/java/uk/ac/ox/ndph/mts/practitioner_service/controller/PractitionerController.java
```diff
@@ -81,7 +82,7 @@ public PractitionerController(EntityService entityService,  SecurityContextUtil
         return ResponseEntity.status(CREATED).body(new Response(roleAssignmentId));
     }
 
-    @PreAuthorize("@authorisationService.authoriseUserRoles(#userIdentity)")
+    @PreAuthorize("@authorisationService.authUserRoles(#userIdentity)")
```
darrengsmithson, line 85: general tidy as I wanted to make the methods less verbose to allow checkstyle to pass for line length

practitioner-service/src/test/java/uk/ac/ox/ndph/mts/practitioner_service/TestAuthorisationConfigurationProvider.java
```diff
@@ -20,9 +20,8 @@ public AuthorisationService authorisationService() {
         var mockService = Mockito.mock(AuthorisationService.class);
         Mockito.when(mockService.authorise(anyString())).thenReturn(true);
         Mockito.when(mockService.authorise(anyString(), anyString())).thenReturn(true);
-        Mockito.when(mockService.authorise(anyString(), anyList(), anyString())).thenReturn(true);
         Mockito.when(mockService.authorise(anyString(), anyList())).thenReturn(true);
-        Mockito.when(mockService.authoriseUserRoles(anyString())).thenReturn(true);
+        Mockito.when(mockService.authUserRoles(anyString())).thenReturn(true);
```
darrengsmithson, line 24: general tidy as I wanted to make the methods less verbose to allow checkstyle to pass for line length

role-service/src/main/java/uk/ac/ox/ndph/mts/role_service/controller/RoleController.java
```diff
@@ -59,7 +59,7 @@ public Role getOne(@PathVariable String id) throws ResponseStatusException {
     }
 
     @GetMapping(params = "ids")
-    @PreAuthorize("@authorisationService.authoriseUserPermissionRoles(#ids)")
+    @PreAuthorize("@authorisationService.authUserPermRoles(#ids)")
```
darrengsmithson, line 62: general tidy as I wanted to make the methods less verbose to allow checkstyle to pass for line length

security/src/main/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationService.java
```diff
@@ -90,56 +69,56 @@ public boolean authorise(String requiredPermission) {
...
-            List<SiteDTO> sites = siteServiceClient.getAssignedSites(
-                                                    SiteServiceClient.bearerAuth(securityContextUtil.getToken()));
+            //If we have required permission at a role and no site given authentication is correct
+            if (siteIds.isEmpty() || siteIds.get(0).isEmpty()) {
+                return true;
+            }
 
-            Set<String> userSites = siteUtil.getUserSites(sites, rolesAssignmentsWithPermission);
+            //Check if the permission exists in roleassignments for siteId
+            if (roleAssnmntsWtPerm.stream().anyMatch(ra -> siteIds.get(0).equalsIgnoreCase(ra.getSiteId()))) {
```
alexbaxter, line 115: This looks to me like it only checks the first site ID in the siteIds parameter? If so need to change the method signature probably (or fix implementation).

security/src/main/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationService.java
```diff
@@ -162,85 +141,33 @@ public boolean authoriseUserRoles(String userIdentityParam) {
...
+        List<RoleAssignmentDTO> roleAssignments = practServClnt.getUserRoleAssignments(getUserId(), getAuthHeaders());
 
         List<String> roleAssignmentIds = roleAssignments.stream()
-                .map(RoleAssignmentDTO::getRoleId).collect(Collectors.toList());
+            .map(RoleAssignmentDTO::getRoleId).collect(Collectors.toList());
 
         return roleAssignmentIds.containsAll(ids);
     }
 
-    /**
```
alexbaxter, line 178: Is this method reimplemented somewhere else? We do need it.

security/src/main/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationService.java
```diff
@@ -162,85 +141,33 @@ public boolean authoriseUserRoles(String userIdentityParam) {
...
-                .map(RoleAssignmentDTO::getRoleId).collect(Collectors.toList());
-
+    private List<RoleAssignmentDTO> getRoleAssmntsWtPerm(String reqPerm,
+                                                         List<RoleAssignmentDTO> roleAssmnts) {
         //get permissions for the the practitioner role assignments
         //and filter role assignments to be only those which have the required permission in them
-        Set<String> rolesWithPermission = roleServiceClient.getRolesByIds(roleIds,
-                RoleServiceClient.bearerAuth(securityContextUtil.getToken())).stream()
-                .filter(roleDto -> hasRequiredPermissionInRole(roleDto, requiredPermission))
+        Page<RoleDTO> roleDTOs = roleServClnt.getPage(0, 500, RoleServiceClient.bearerAuth(getToken()));
```
alexbaxter, line 163: If this is paged then need to handle paging to get all the roles, don't assume there is some limit.

site-service-client/src/main/java/uk/ac/ox/ndph/mts/siteserviceclient/SiteServiceClient.java
```diff
@@ -79,11 +79,11 @@ public SiteResponseDTO createEntity(final SiteDTO site,
         return Arrays.asList(requestExecutor.sendBlockingGetRequest(webClient, uri, SiteDTO[].class, authHeaders));
     }
 
-    public List<SiteDTO> getAssignedSites(final Consumer<HttpHeaders> authHeaders) {
+    public List<String> getParentSiteIds(final String siteId, final Consumer<HttpHeaders> authHeaders) {
```
alexbaxter, line 82: Is getAssignedSites() implemented somewhere? Again, this is a method we do need.

alexbaxter (commented):
The main change to me looks like some methods have been moved from the auth-service to the site-service, presumably for efficiency; auth-service still depends on site-service but it did anyway.

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/controller/SiteController.java
```diff
@@ -37,22 +38,20 @@ public SiteController(SiteService siteService) {
...
+    public List<SiteDTO> getSites(@RequestParam(value = "role", required = false) String role) {
+        return siteService.findSites();
     }
 
     @GetMapping("/{id}")
     public ResponseEntity<Site> findSiteById(@PathVariable String id) {
         return ResponseEntity.ok(siteService.findSiteById(id));
     }
 
+    @GetMapping("/parents/{id}")
```
KateSant, line 52: If I understand this correctly, it's deliberate that this method does not have any authz on it and important, since it is called by the authz component and we otherwise get into circularity?  If so should there be a comment here to tell future developers that this method is special?

security/src/main/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationService.java
```diff
@@ -90,56 +69,56 @@ public boolean authorise(String requiredPermission) {
...
-            var rolesAssignmentsWithPermission = getRolesAssignmentsWithPermission(requiredPermission, roleAssignments);
-
-            if (rolesAssignmentsWithPermission.isEmpty()) {
+            //Check for the permission in role assignments
+            var roleAssnmntsWtPerm = getRoleAssmntsWtPerm(requiredPerm, roleAssnmnts);
+            if (roleAssnmntsWtPerm.isEmpty()) {
                 return false;
             }
 
-            List<SiteDTO> sites = siteServiceClient.getAssignedSites(
```
KateSant, line 137: Have I understood this correctly?  This is the bit of logic where the authz component engages with the site tree hierarchy.   It is needed because permissions flow down hierarchically.   The authz component gets the data for this operation by making a GET request to site service.  Previously the implementation was a call to the /sites/assigned method giving all sites assigned to the user.  The new implementation is a call to /sites/parents/id method to get all parents of a child site.    In both the new and the old implementation, the site service endpoint has to be unauthorized to avoid loops.  Have I got it?

security/src/main/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationService.java
```diff
@@ -90,56 +69,56 @@ public boolean authorise(String requiredPermission) {
...
-     * @param requiredPermission the required permission
-     * @param entitiesSiteIds list of site ids to check if the user is authorised on them
+     * @param requiredPerm the required permission
+     * @param siteIds list of site ids to check if the user is authorised on them
      * @return true if authorised - has the required permission and is authorised on all sites in entitiesSiteIds
      */
-    public boolean authorise(String requiredPermission, List<String> entitiesSiteIds)  {
-
-
+    public boolean authorise(String requiredPerm, List<String> siteIds)  {
```
KateSant, line 76: Nothing to do with your refactor Darren just an extra thought.   This method makes GET requests to practitioner service, site service and role service, to get the data it needs to do this logic.   I think it's unavoidable (leaving aside caching for now).    But it's a bit hidden to the uninitiated.  Would it be useful to future developers to add a big comment, flagging that this method does all this traffic?

security/src/main/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationService.java
KateSant, line 163: Why the switch from getting only the limited list of roles, to getting all roles and then filtering afterwards?

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/controller/SiteController.java
```diff
@@ -37,22 +38,20 @@ public SiteController(SiteService siteService) {
...
 
-    @PreAuthorize("@authorisationService.authorise('view-site')")
-    @PostAuthorize("@authorisationService.filterUserSites(returnObject.getBody(), #role)")
+    @PostAuthorize("@siteServiceImpl.filterUserSites(returnObject, #role, 'view-site')")
     @GetMapping
-    public ResponseEntity<List<Site>> getSites(@RequestParam(value = "role", required = false) String role) {
-        return ResponseEntity.status(HttpStatus.OK).body(siteService.findSites());
-    }
-
-    @GetMapping("/assigned")
```
KateSant, line 47: I like that the new implementation removes the need for this endpoint, which gave site service dependent on practitioner service, was hard to understand and generally yuk.   The new /sites/parents/id endpoint is clearer in my view.

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/service/SiteServiceImpl.java
```diff
@@ -116,11 +143,16 @@ public String save(final Site site) {
...
-        final List<Site> sites = this.siteStore.findAll();
+    public List<SiteDTO> findSites() {
+        final List<SiteDTO> sites = this.siteStore.findAll().stream()
+            .map(site -> convertSite(site))
+            .collect(Collectors.toList());
         if (sites.isEmpty()) {
             throw new InvariantException(Services.NO_ROOT_SITE.message());
         }
+
+        populateParentName(sites);
```
KateSant, line 154: This is the performance improvement that saves the UI from having to come back with n+1 little requests for parentnames?  I think this is a good idea.

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/service/SiteServiceImpl.java
```diff
@@ -163,5 +200,124 @@ Site findRootSite() throws ResponseStatusException {
...
+    }
+
+    private SiteAddressDTO convertAddress(SiteAddress siteAddress) {
+        return new SiteAddressDTO(siteAddress.getAddress1(), siteAddress.getAddress2(), siteAddress.getAddress3(),
+            siteAddress.getAddress4(), siteAddress.getAddress5(), siteAddress.getCity(), siteAddress.getCountry(),
+            siteAddress.getPostcode());
+    }
+
+    @Override
+    public boolean filterUserSites(List<SiteDTO> sitesReturnObject,
```
KateSant, line 225: I think this method was like this before you moved it Darren so not your changes, but does it break the SRP by doing 2 things (1) check authz returning true/false (2) filtering sites ?  Also it seems to operate on the sitesReturnObject argument - side effect?

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/service/SiteServiceImpl.java
```diff
@@ -163,5 +200,124 @@ Site findRootSite() throws ResponseStatusException {
...
+                .contains(site.getSiteId()));
+    }
+
+    /**
+     * Validate if the role assignments have the required permission linked to them.
+     * @param reqPerm action required permission
+     * @param roleAssmnts user role assignments
+     * @return true if required permission is present in one of the roles
+     */
+    private List<RoleAssignmentDTO> getRoleAssmntsWtPerm(String reqPerm,
```
KateSant, line 292: Is getRoleAssmntsWtPerm the duplicated method that we discussed earlier?  Needed in both site-service and the authz component?   I am still not sure whether it's better to start an mts-utilities library (risking creeping back towards a monolith) or to relax about duplication.   Just recording our thoughts here for ref.

security/src/main/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationService.java
darrengsmithson: It was with 1 eye on future clean-up/refactor/maintenance - I think that we have an endpoint (roleServiceClient.getRolesByIds) created just for this scenario - I couldnt see it anywhere else - I thought why use it when we can just get all roles - itll never be a perf issue as there wont be that many roles in the DB. We can then remove the endpoint at a later date when were sure that it wont be used - another instance where I didnt want to refactor too much at the mo but made changes to help future refactoring a little easier. If there is a legit reason for the roleServiceClient.getRolesByIds endpoint other than this Im happy to put it back

security/src/main/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationService.java
darrengsmithson: Hi Alex - after looking into the logic around using a List of siteIds rather than a siteId string - the code is written at the mo to check that if a siteId is null then throw an access is denied error unless its from the initservice which is determied before the error is thrown. Currently we have a create site endpoint that passes the parentSiteId as the siteId, the assign role endpoint that passes the siteId for the user to be assigned to and a host of other endpoints that dont pass any siteId. If any endpoints that dont pass the siteId need to be authzd we create an empty list of siteIds to bypass the null check for the endpoints that should pass a siteId....all this is a little messy to me and also highlights the reason why auth service shouldnt be executing business rules for all services...anyhow if we are to change the current approach  a separate PR will be needed.

security/src/main/java/uk/ac/ox/ndph/mts/security/authorisation/AuthorisationService.java
shajidissa: @darrengsmithson put a TODO here ? wdyt

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/controller/SiteController.java
```diff
@@ -37,22 +38,20 @@ public SiteController(SiteService siteService) {
...
-    @PostAuthorize("@authorisationService.filterUserSites(returnObject.getBody(), null)")
```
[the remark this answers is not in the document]
shajidissa: Assuming UI needs to be changed as it calls this ?

shajidissa (approved):

KateSant (approved):

site-service/src/main/java/uk/ac/ox/ndph/mts/site_service/service/SiteServiceImpl.java
KateSant: Is it worth disentangling them?  Or -halfway house - returning a Pair to make clear that there are two things going on?   Not a blocker - have approved.

